PR #127, opened 2021-11-10
CASMPET-5145

leliasen-hpe:
## Summary and Scope

If a postgres cluster has lag greater than the maximum value allowed, then the test will wait and try again for a specified number of attempts. This is an enhancement of the existing script.

The change is backwards compatible.

## Testing

### Tested on:

  * Drax
  * Virtual Shasta

### Test description:

- Were the install/upgrade-based validation checks/tests run (goss tests/install-validation doc)?
- Were continuous integration tests run? If not, why?
- Was upgrade tested? If not, why?
- Was downgrade tested? If not, why?
- Were new tests (or test issues/Jiras) created for this change?

This performs read only operations.

## Pull Request Checklist

- [ ] Copyrights updated
- [ ] License file intact
- [x] Target branch correct
- [ ] CHANGELOG.md updated


bklei (approved):

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -1,19 +1,56 @@
...
-         \?) echo "usage: postgres_replication_lag.sh           # Only print 'PASS' upon success"
+         \?) echo "usage: postgres_replication_lag.sh           # Print 'OK' and 'PASS' upon success"
              echo "       postgres_replication_lag.sh -p        # Print all results and errors if found. Use for manual check."
+             echo "       postgres_replication_lag.sh -p -m <max_allowed_lag> -a <number_attempts> -w <wait_seconds_between_attempts>"
+             echo "default parameters: -m 0  -a 10  -w 10"
              exit 3;;
     esac
 done
 
+# The POSTGRES_MAX_LAG may be exported by the user to control
```
brantk-hpe, line 24: "The POSTGRES_MAX_LAG may" -> "The POSTGRES_MAX_LAG environment variable may"

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -1,19 +1,56 @@
...
+# the maximum lag value permitted by this script. Setting its value to
+# a negative number or a non-integer value has the effect of skipping the
+# maximum lag check. In other words, if one wishes to skip this check,
+# one could:
+# export POSTGRES_MAX_LAG=skip
+POSTGRES_MAX_LAG=${POSTGRES_MAX_LAG:-'0'}
+
+# POSTGRES_MAX_ATTEMPTS specifies the maximum number of times the
+# postgres check will be performed on a given cluster before failing. Note that
```
brantk-hpe, line 33: "postgres" -> "PostgreSQL" -- PostgreSQL is another company's trademark we should use it as they request.

This applies throughout this script.

kimjensen-hpe (approved):

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -1,19 +1,56 @@
...
+# failures other than due to maximum lag are always fatal and are not retried.
+# If unset or set to a non-positive integer, default to 10
+if [[ ! $POSTGRES_MAX_ATTEMPTS =~ ^[1-9][0-9]*$ ]]; then
```
brantk-hpe, line 36: Looks like the number can't start with 0? (For example 010)

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -1,19 +1,56 @@
...
+# If unset or set to a non-positive integer, default to 10
+if [[ ! $POSTGRES_WAIT_SECONDS_BETWEEN_ATTEMPTS =~ ^[1-9][0-9]*$ ]]; then
+    POSTGRES_WAIT_SECONDS_BETWEEN_ATTEMPTS=10
+fi
+
+if [[ ! $POSTGRES_MAX_LAG =~ ^[0-9][0-9]*$ ]]; then
+    echo "Skipping postgres cluster max lag checks because of POSTGRES_MAX_LAG setting"
+    exit 0
+else
+    echo "Postgres cluster checks may take several minutes, depending on latency"
```
brantk-hpe, line 51: latency of what?

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -25,19 +62,54 @@ do
...
 
-    for lag in $(kubectl -n $c_ns exec $first_member -- patronictl list --format json 2>/dev/null | jq '.[]."Lag in MB"')
-    do
-        if [[ $lag != "\"\"" ]] && [[ $lag == "\"unknown"\" || $lag -gt 50 ]]; 
-        then 
-            if [[ $print_results -eq 1 ]]; 
-            then 
-                echo "Warning: $c_name has Lag: $lag"; 
-                failFlag=1 
+    echo -n "$c_name - "
```
brantk-hpe, line 65: Shouldn't this echo only happen if -p was used?

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -25,19 +62,54 @@ do
...
+        c_unknown_lag=$(echo $cluster_lag | grep "unknown" | wc -l)
+        if [[ -n $c_lag_history ]]; then
+            c_lag_history="${c_lag_history}, $c_max_lag"
+        else
+            c_lag_history="$c_max_lag"
+        fi
+
+        #check lag:unknown
+        if [[ $c_unknown_lag -gt 0 ]]; then
+            echo -e "\n--- ERROR --- $c_name cluster has lag: unknown"
```
brantk-hpe, line 87: Isn't this only supposed to be printed if the -p option was used? I'm confused as to what the difference is between using -p and not using -p.

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -25,19 +62,54 @@ do
...
+            if [[ $print_results -eq 1 ]]
+            then failFlag=1
                 kubectl -n $c_ns exec $first_member -- patronictl list 2>/dev/null
+                break;
             else exit 1; fi
         fi
+
+        if [[ $POSTGRES_MAX_LAG =~ ^[0-9][0-9]*$ ]]; then
```
brantk-hpe, line 95: This script would have exited at line 49 if `! $POSTGRES_MAX_LAG =~ ^[0-9][0-9]*$` so this check seems redundant.

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -1,19 +1,56 @@
...
              echo "       postgres_replication_lag.sh -p        # Print all results and errors if found. Use for manual check."
+             echo "       postgres_replication_lag.sh -p -m <max_allowed_lag> -a <number_attempts> -w <wait_seconds_between_attempts>"
+             echo "default parameters: -m 0  -a 10  -w 10"
              exit 3;;
-         \?) echo "usage: postgres_replication_lag.sh           # Only print 'PASS' upon success"
+         \?) echo "usage: postgres_replication_lag.sh           # Print 'OK' and 'PASS' upon success"
              echo "       postgres_replication_lag.sh -p        # Print all results and errors if found. Use for manual check."
+             echo "       postgres_replication_lag.sh -p -m <max_allowed_lag> -a <number_attempts> -w <wait_seconds_between_attempts>"
+             echo "default parameters: -m 0  -a 10  -w 10"
              exit 3;;
```
brantk-hpe, line 20: Was this supposed to be changed to exit 4 since the FAIL case now uses exit 3?

brantk-hpe (changes requested):
We need to use company's trademarks correctly otherwise they might sue us.

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: fixed

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: Changed to "PostgreSQL"

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: Reworded

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: I changed this to only print when -p is used. I also added another option, -e, which will exit the script on failure. The options -p -e can be used together to print OK and failures but will exit the script once it fails.

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: Yes, changed to only print with -p

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: Removed check

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: changed to exit 4

goss-testing/scripts/postgres_replication_lag.sh
```diff
@@ -1,43 +1,119 @@
...
+# maximum lag check. In other words, if one wishes to skip this check,
+# one could:
+# export POSTGRES_MAX_LAG=skip
+POSTGRES_MAX_LAG=${POSTGRES_MAX_LAG:-'0'}
+
+# POSTGRES_MAX_ATTEMPTS specifies the maximum number of times the
+# PostgreSQL check will be performed on a given cluster before failing. Note that
+# failures other than due to maximum lag are always fatal and are not retried.
+# If unset or set to a non-positive integer, default to 10
+if [[ ! $POSTGRES_MAX_ATTEMPTS =~ ^[0-9][0-9]*$ ]]; then
```
brantk-hpe, line 40: this regular expression matches 0, so I don't think this does what the documentation above says since 0 is a non-positive integer.

brantk-hpe (dismissed):

goss-testing/scripts/postgres_replication_lag.sh
leliasen-hpe: I changed this to` [[ ! $POSTGRES_MAX_ATTEMPTS =~ ^[1-9][0-9]*$ ]]`. We do not want 0 to be a valid input because the test will not run if the input is 0. I also think it is okay to assume the number entered will not start with a 0.

kimjensen-hpe (dismissed):

bklei (approved):

bklei (approved):

PR #141, opened 2021-11-18
CASMINST-3587 1.2 : TEST - Fanta - preflight check failed - remove…

kimjensen-hpe:
## Summary and Scope
Removes tests that fails as part of the upgrade from 1.0 given that these interfaces are not setup until csm-1.2

## Issues and Related PRs

* Resolves [CASMINST-3587](https://connect.us.cray.com/jira/browse/CASMINST-3587)
* Change will also be needed in `master` - completed
* Future work required by NA
* Documentation changes required NA
* Merge with/before/after NA

## Testing
NA

## Risks and Mitigations
NA

## Pull Request Checklist

- [ ] Copyrights updated
- [ ] License file intact
- [x] Target branch correct
- [ ] CHANGELOG.md updated


brantk-hpe (approved):

goss-testing/suites/ncn-upgrade-preflight-tests.yaml
```diff
@@ -4,11 +4,8 @@ gossfile:
   ../tests/goss-k8s-nodes-ready.yaml: {}
   ../tests/goss-command-available.yaml: {}
   ../tests/goss-basecamp-json-ncns.yaml: {}
-  ../tests/goss-network-interfaces.yaml: {}
```
dzou-hpe, line 7: do we not want to run these checks? or because we have BI-CAN checks in these tests now so we can't run these any more before upgrade?

dzou-hpe (approved):

rgross727 (approved):

goss-testing/suites/ncn-upgrade-preflight-tests.yaml
kimjensen-hpe: This was removed since the following will not pass until csm -1.2 is installed
```
common/goss-network-interfaces.yaml:      desc: Ensures VLAN was created for network interface '{{$interface}}'.
```
For upgrades from csm-1.2 we could add this test back but for upgrades from csm-1.0, my understanding is that we want to remove it.

PR #216, opened 2022-03-03
CASMINST-4091: Add logging for some goss tests

mharding-hpe:
## Summary and Scope

It can currently be difficult to debug failures of goss tests, particularly when they are intermittent, because the results include minimal information about the failure.

This PR creates a new script (log_run.sh) which is a wrapper script used when calling a command or test script. This wrapper logs what is being run, its stdout, its stderr, and its return code to a log file on the system. It is transparent to the caller, however -- it passes through all of stdout to stdout, all of stderr to stderr, and the return code of the command being run. The wrapper script itself generates no output except for usage errors if invalid arguments are used with it.

I modified several tests to make use of this wrapper script. There are many other candidates where it can be used, but I started with the lowest hanging fruit.

This PR also includes some minor fixes and enhancements to other tests, as well as increasing the timeouts of several tests (based on false failures I observed due to timeouts of these tests).

## Issues and Related PRs

None

## Testing

I tested on the PIT node on wasp, and on a master, worker, and storage node on hela. I also did testing of the wrapper script itself in isolation.

## Risks and Mitigations

Low risk. I have not found any cases where the wrapper script altered the behavior of any of the tests.

## Pull Request Checklist

- [X] Copyrights updated
- [X] License file intact
- [X] Target branch correct
- [N/A] CHANGELOG.md updated
- [N/A] [HPC Product Announcement](https://cray.slack.com/archives/C026TVCSXLH) prepared, if applicable


mharding-hpe:
To simplify the code review process, I'm going to decline my current PR and break it into smaller ones. As it is I think the prospect of review may be daunting to some.

mharding-hpe:
Reduced code PR (adding logging to only 2 tests) is here:
https://github.com/Cray-HPE/csm-testing/pull/220

PR #222, opened 2022-03-08
CASMINST-4161 adjust interface mac bond check test for optional can0

jacobsalmela:
Signed-off-by: Jacob Salmela <jacob.salmela@hpe.com>

## Summary and Scope

- Adjusts `check_interface_mac_matches_bond.sh` to accept multiple arguments to facilitate checking them all in the same script and removes logging in favor of the bullet below
- Adjusts `goss-interfaces-have-same-mac.yaml` to use `log_run.sh`
- Adjusts `goss-interfaces-have-same-mac.yaml` to utilize a small loop while retaining functionality of the `.Env.GOSS_BASE` variable since that variable cannot be evaluated in the current loop of `network_interfaces`.
- Fixes a mismatched quote in `ncn_healthcheck` (opening `"` with a closing `'`) introduced with #220 

## Issues and Related PRs

- CASMINST-4161
- Relates to CASMINST-4183

## Testing

Tested manually

### Tested on:

  * `surtur`
  * `wasp`

### Test description:

on wasp **without** `can0`

```
wasp-ncn-m001-pit:~ # goss --vars /opt/cray/tests/install/ncn/vars/variables-ncn.yaml validate
.

Total Duration: 0.043s
Count: 1, Failed: 0, Skipped: 0
wasp-ncn-m001-pit:~ # cat /opt/cray/tests/install/logs/check_mac/20220308_202347_297618660_22208.log
log_run argument(s): -l check_mac
Executable: '/opt/cray/tests/install/ncn//scripts/check_interface_mac_matches_bond.sh'
4 argument(s) to executable: 'bond0.nmn0' 'bond0.hmn0' 'bond0.can0' 'bond0.cmn0'

## 20220308_202347_306114781 ##                      executable starting ##
###########################################################################
Checking that MAC address of interface 'bond0.nmn0' matches MAC address of bond0...

b8:59:9f:f9:1b:fe (bond0)
b8:59:9f:f9:1b:fe (bond0.nmn0)
Checking that MAC address of interface 'bond0.hmn0' matches MAC address of bond0...

b8:59:9f:f9:1b:fe (bond0)
b8:59:9f:f9:1b:fe (bond0.hmn0)
Checking that MAC address of interface 'bond0.can0' matches MAC address of bond0...

Device "bond0.can0" does not exist.
bond0.can0 is optional...skipping
###########################################################################
## 20220308_202347_331642266 ##       executable completed (exit code=0) ##
```

on surtur **with** `can0`

```
[...]

PR #226, opened 2022-03-09
CASMINST-4196 adjust vlan test to ignore options bond0.can0 interface

jacobsalmela:

## Summary and Scope

Adjusts a test to ignore `bond0.can0` as it is now optional.

## Issues and Related PRs

* Resolves CASMINST-4196
* Relates to CASMINST-4195
* Relates to #222 

## Testing

  * `wasp`

### Test description:

```
ncn-m001:~ # export GOSS_BASE=/opt/cray/tests/install/ncn/
ncn-m001:~ # export GOSS_FILE=/opt/cray/tests/install/ncn/tests/goss-network-interfaces.yaml
ncn-m001:~ # goss --vars /opt/cray/tests/install/ncn/vars/variables-ncn.yaml validate
....

Total Duration: 0.014s
ncn-m001:~ # echo $?
0
ncn-m001:~ #
```

## Risks and Mitigations

Low

## Pull Request Checklist

- [ ] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [ ] CHANGELOG.md updated

### DIFF
--- a/goss-testing/tests/common/goss-network-interfaces.yaml
+++ b/goss-testing/tests/common/goss-network-interfaces.yaml
@@ -24,6 +24,7 @@
 # Variable set: network_interfaces
 command:
 {{range $interface := index .Vars "network_interfaces"}}
+  {{if ne $interface "bond0.can0"}}
   interface_{{$interface}}:
     title: Interface '{{$interface}}' Has VLAN
     meta:
@@ -33,4 +34,5 @@ command:
     exit-status: 0
     timeout: 10000
     skip: false
+  {{end}}
 {{end}}


goss-testing/tests/common/goss-network-interfaces.yaml
```diff
@@ -24,12 +24,17 @@
...
       desc: Ensures VLAN was created for network interface '{{$interface}}'.
       sev: 0
-    exec: 'ip link show dev {{$interface}}'
+    exec: |-
+        if ! eval 'ip link show dev {{$interface}}' 1>/dev/null; then
+          if [[ "{{$interface}}" == "bond0.can0" ]]; then
+            echo "{{$interface}}is optional...skipping"
+          fi
+        fi
     exit-status: 0
```
mharding-hpe, line 38: The above script will always exit 0, I think. Or am I missing something? I don't see how this test would fail.

mharding-hpe (changes requested):

goss-testing/tests/common/goss-network-interfaces.yaml
mharding-hpe: Like with the other PR, though, if you just want to skip the test for can0, using goss conditional execution could be simpler:
```
command:
{{range $interface := index .Vars "network_interfaces"}}
    {{if ne {{$interface}} "bond0.can0"}}
        interface_{{$interface}}_has_vlan:
            title: Interface '{{$interface}}' Has VLAN
            meta:
                desc: Ensures VLAN was created for network interface '{{$interface}}'.
                sev: 0
            exec: ip link show dev {{$interface}}
            exit-status: 0
            timeout: 10000
            skip: false
    {{end}}
{{end}}
```

mharding-hpe (changes requested):

goss-testing/tests/common/goss-network-interfaces.yaml
jacobsalmela: Fixed up.  Thanks!

jacobsalmela: Actually, I was just thinking this will always skip the can.  If it is there, we do want to check it... we might need to re-think this a bit

jacobsalmela: OK I updated the logic here to fail gracefully if can0 is not there:
```
    {{if eq $interface "bond0.can0"}}
    # it's ok if the bond0.can0 fails since it is optional
    exec: "if ! eval ip -4 a s dev {{$interface}} | awk '/inet / {print $2}'; then : ; fi"
    {{else}}
    exec: "ip -4 a s dev {{$interface}} | awk '/inet / {print $2}'"
    {{end}}
```

goss-testing/tests/common/goss-network-interfaces.yaml
```diff
@@ -29,7 +29,13 @@ command:
     meta:
       desc: Ensures VLAN was created for network interface '{{$interface}}'.
       sev: 0
-    exec: 'ip link show dev {{$interface}}'
+    # For bond0.can0, if it does not exist, we do not fail. Echo a fake IP so the test passes
+    exec: |-
```
mharding-hpe, line 33: This has the same effect as just not running the test for bond0.can0, since it will never fail for that interface. If that is the desired behavior, may as well just skip the test for bond0.can0 using goss conditional logic, IMO.

mharding-hpe (changes requested):

goss-testing/tests/common/goss-network-interfaces.yaml
jacobsalmela: Forgot to push here.  Check now

mharding-hpe: This still will never fail for bond0.can0

mharding-hpe (changes requested):

goss-testing/tests/common/goss-network-interfaces.yaml
mharding-hpe: I would change it to something like:
```
command:
    {{range $interface := index .Vars "network_interfaces"}}
        {{if ne {{$interface}} "bond0.can0"}}
            check_interface_vlan_{{$interface}}:
                title: Interface '{{$interface}}' Has VLAN
                meta:
                    desc: Ensures VLAN was created for network interface '{{$interface}}'.
                    sev: 0
                exec: 'ip link show dev {{$interface}}'
                exit-status: 0
                timeout: 10000
                skip: false
        {{end}}
    {{end}}
```

jacobsalmela: That seems to work!

jacobsalmela: Fixed

goss-testing/tests/common/goss-network-interfaces.yaml
```diff
@@ -22,8 +22,33 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 # Variable set: network_interfaces
+#
```
mharding-hpe, line 25: It looks like you added in another license header

mharding-hpe (changes requested):

goss-testing/tests/common/goss-network-interfaces.yaml
jacobsalmela: Fixed ☹️

mharding-hpe (approved):

gbaker-hpe (approved):

PR #227, opened 2022-03-09
CASMINST-4195 adjust ip test to ignore options bond0.can0 interface

jacobsalmela:

## Summary and Scope

Adjusts a test to ignore `bond0.can0` as it is now optional.

## Issues and Related PRs

* Resolves CASMINST-4195
* Relates to CASMINST-4196
* Relates to #222 

## Testing

  * `wasp`

### Test description:

```
ncn-m001:~ # export GOSS_BASE=/opt/cray/tests/install/ncn/
ncn-m001:~ # export GOSS_FILE=/opt/cray/tests/install/ncn/tests/goss-interface-has-ip.yaml
ncn-m001:~ # goss --vars /opt/cray/tests/install/ncn/vars/variables-ncn.yaml validate
....

Total Duration: 0.013s
ncn-m001:~ #
```

## Risks and Mitigations

Low

## Pull Request Checklist

- [ ] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [ ] CHANGELOG.md updated

### DIFF
--- a/goss-testing/tests/common/goss-interface-has-ip.yaml
+++ b/goss-testing/tests/common/goss-interface-has-ip.yaml
@@ -29,10 +29,18 @@ command:
     meta:
       desc: Validates that interface '{{$interface}}' has an IP address
       sev: 0
-    exec: "ip addr show dev {{$interface}} scope global | grep 'inet ' | awk '{print $2}'"
+    # For bond0.can0, if it does not exist, we do not fail. Echo a fake IP so the test passes
+    exec: |-
+      if ip link show dev {{$interface}}; then
+        ip -4 a s dev {{$interface}} | awk '/inet / {print $2}'
+      fi
     exit-status: 0
     stdout:
     - /[1-9]/
     timeout: 10000
+    {{if eq $interface "bond0.can0"}}
+    skip: true
+    {{else}}
     skip: false
+    {{end}}
 {{end}}


goss-testing/tests/common/goss-interface-has-ip.yaml
```diff
@@ -29,10 +29,16 @@ command:
...
+      if ! eval ip addr show dev {{$interface}} 1>/dev/null; then
+        if [[ "{{$interface}}" == "bond0.can0" ]]; then
+          echo "{{$interface}} is optional...skipping"
+        else
+          ip addr show dev {{$interface}} scope global | grep 'inet ' | awk '{print $2}'
+        fi
+      fi
     exit-status: 0
     stdout:
-    - /[1-9]/
```
mharding-hpe, line 35: If you're removing this line, how are you detecting failures? The above script is always going to exit with a 0 return code, unless I'm missing something.

mharding-hpe (changes requested):

goss-testing/tests/common/goss-interface-has-ip.yaml
mharding-hpe: If you just want to skip the test for that interface, you could use conditional execution in goss. Something like this:
```
command:
{{range $interface := index .Vars "network_interfaces"}}
    {{if ne {{$interface}} "bond0.can0"}}
        check_interface_ip_{{$interface}}:
            title: Interface '{{$interface}}' Has IP
            meta:
                desc: Validates that interface '{{$interface}}' has an IP address
                sev: 0
            exec: |-
                set -eo pipefail
                ip addr show dev {{$interface}} scope global | grep inet' | awk '{print $2}'
            exit-status: 0
            stdout:
                - /[1-9]/
            timeout: 10000
            skip: false
    {{end}}
{{end}}
```

mharding-hpe (changes requested):

goss-testing/tests/common/goss-interface-has-ip.yaml
jacobsalmela: Oh slick. Must have overlooked that capability in the docs.  Makes sense though

jacobsalmela: I think I mistakenly deleted that stdout..I'll fix this up like the other one

jacobsalmela: Actually, I was just thinking this will always skip the can.  If it is there, we do want to check it... we might need to re-think this a bit

mharding-hpe: It's easy to overlook stuff in the goss docs, I think. I don't find them particularly well organized. :) 

In that case, I think you will need to have the logic in the test script itself. I don't know a way in goss to run a command on the system and, based on its output or return code, conditionally execute a test or not. So I think you will need to do something like your solution.

jacobsalmela: OK I updated the logic here to fail gracefully if can0 is not there:
```
    {{if eq $interface "bond0.can0"}}
    # it's ok if the bond0.can0 fails since it is optional
    exec: "if ! eval ip -4 a s dev {{$interface}} | awk '/inet / {print $2}'; then : ; fi"
    {{else}}
    exec: "ip -4 a s dev {{$interface}} | awk '/inet / {print $2}'"
    {{end}}
```

goss-testing/tests/common/goss-interface-has-ip.yaml
```diff
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2014-2022 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
```
mharding-hpe, line 4: Should you be dropping the previous years?

goss-testing/tests/common/goss-interface-has-ip.yaml
```diff
@@ -21,6 +21,28 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 #
+# (C) Copyright 2014-2022 Hewlett Packard Enterprise Development LP.
```
mharding-hpe, line 24: And it looks like you are adding an additional license header to the file?

goss-testing/tests/common/goss-interface-has-ip.yaml
```diff
@@ -29,10 +51,54 @@ command:
     meta:
       desc: Validates that interface '{{$interface}}' has an IP address
       sev: 0
-    exec: "ip addr show dev {{$interface}} scope global | grep 'inet ' | awk '{print $2}'"
+    # it's ok if the bond0.can0 fails since it is optional
+    exec: |-
+    #
+# MIT License
```
mharding-hpe, line 57: And another header? I'm very confused...

mharding-hpe (approved):

goss-testing/tests/common/goss-interface-has-ip.yaml
mharding-hpe: Need to clear up the license header stuff. :)

mharding-hpe (changes requested):

goss-testing/tests/common/goss-interface-has-ip.yaml
```diff
@@ -24,15 +24,26 @@
 # Variable set: network_interfaces
 command:
 {{range $interface := index .Vars "network_interfaces"}}
-  check_interface_ip_{{$interface}}:
-    title: Interface '{{$interface}}' Has IP
+  interface_{{$interface}}:
+    title: Interface '{{$interface}}' Has VLAN
     meta:
-      desc: Validates that interface '{{$interface}}' has an IP address
+      desc: Ensures VLAN was created for network interface '{{$interface}}'.
```
mharding-hpe, line 30: This test is the IP address test, not the VLAN test, so the previous description was correct.

mharding-hpe (changes requested):

goss-testing/tests/common/goss-interface-has-ip.yaml
```diff
@@ -24,15 +24,26 @@
 # Variable set: network_interfaces
 command:
 {{range $interface := index .Vars "network_interfaces"}}
-  check_interface_ip_{{$interface}}:
-    title: Interface '{{$interface}}' Has IP
```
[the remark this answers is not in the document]
jacobsalmela: Can't copy paste...

goss-testing/tests/common/goss-interface-has-ip.yaml
jacobsalmela: Fixed

mharding-hpe (approved):

heemstra (approved):

gbaker-hpe (dismissed):
Is there a way that we can disable part of the test instead of mocking a fake IP?

jacobsalmela:
@gbaker-hpe possibly.  Let me look.  I know not all keys are required, so we may be able to skip one of the output checks...

jacobsalmela:
Sweet.

```
ncn-m001:~ # GOSS_BASE=/opt/cray/tests/install/ncn/ goss --gossfile /opt/cray/tests/install/ncn/tests/goss-interface-has-ip.yaml --vars /opt/cray/tests/install/ncn/vars/variables-ncn.yaml validate
SS......

Failures/Skipped:

Title: Interface 'bond0.can0' Has IP
Meta:
    desc: Validates that interface 'bond0.can0' has an IP address
    sev: 0
Command: check_interface_ip_bond0.can0: exit-status: skipped
Command: check_interface_ip_bond0.can0: stdout: skipped

Total Duration: 0.010s
Count: 8, Failed: 0, Skipped: 2
ncn-m001:~ #
```

mharding-hpe:
> Is there a way that we can disable part of the test instead of mocking a fake IP?

Note that the fake IP isn't visible to any users running the test -- they just see that it passed.

goss-testing/tests/common/goss-interface-has-ip.yaml
```diff
@@ -29,10 +29,20 @@ command:
     meta:
       desc: Validates that interface '{{$interface}}' has an IP address
       sev: 0
-    exec: "ip addr show dev {{$interface}} scope global | grep 'inet ' | awk '{print $2}'"
+    # For bond0.can0, if it does not exist, we do not fail. Echo a fake IP so the test passes
+    exec: |-
+      if ip link show dev {{$interface}}; then
+        ip -4 a s dev {{$interface}} | awk '/inet / {print $2}'
+      elif [[ {{$interface}} == "bond0.can0" ]]; then
```
mharding-hpe, line 36: You should remove this entire "elif" clause now that you added the conditional around the skip statement

mharding-hpe (changes requested):

goss-testing/tests/common/goss-interface-has-ip.yaml
jacobsalmela: Fixed

mharding-hpe (approved):

PR #254, opened 2022-03-25
Bug/casminst-4000-csm-1.2

dle-hpe:
## Summary and Scope

_Summarize what has changed. Explain why this PR is necessary. What is impacted? Is this a new feature, critical bug fix, etc?_
-  improved detection of csm-1.0 and csm-1.2 bgp tests
- support running test during an upgrade

_Is this change backwards incompatible, backwards compatible, or a backwards compatible bugfix?_
- yes
## Issues and Related PRs

* Resolves 
  - CASMINST-4000
  -  CASMINST-4284
  
## Testing

### Tested on:

- hela
- mug
- fanta

### Test description:

- ran script minus the automation to verify results

- Were the install/upgrade-based validation checks/tests run (goss tests/install-validation doc)?n
- Were continuous integration tests run? If not, why?n
- Was upgrade tested? If not, why?n
- Was downgrade tested? If not, why?n
- Were new tests (or test issues/Jiras) created for this change?n

## Risks and Mitigations

_Are there known issues with these changes? Any other special considerations?_

- n/a
## Pull Request Checklist

- [ ] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [ ] CHANGELOG.md updated


goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,46 @@ function err_exit
     exit $rc
 }
 
-kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE ||
-    err_exit 10 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE"
+TOKEN=$(curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret=`kubectl get secrets admin-client-auth -o jsonpath='{.data.client-secret}' | base64 -d` https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token')
+
+# check if metalLB configmap exists
+kubectl -n metallb-system get cm metallb 2>&1 |grep -i error|wc -l > $TMPFILE ||
```
johren-hpe, line 55: When I run this on a 1.2 system, this the script fails on this line because the "grep -i error" returns 1.

And when I run this on a 1.0 system, the script fails on this line because the kubectl -n metallb-system get cm metallb 2>&1 part returns 1.

The set -o pipefail is what causes each fail and trigger the "err_exit" before it gets to the wc -l.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,46 @@ function err_exit
...
+    err_exit 10 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb 2>&1 |grep -i error|wc -l " >> $TMPFILE
```
johren-hpe, line 56: The "ERROR" here is unnecessary because the err_exit already prepends and "ERROR".  This is what I get on the error:

"ERROR: ERROR: Command failed (rc=1): kubectl -n metallb-system get cm metallb 2>&1 |grep -i error|wc -l "

johren-hpe (changes requested):

goss-testing/scripts/check_bgp_neighbors_established.sh
johren-hpe: It looks like lines 55 and 56 just shouldn't be there.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,46 @@ function err_exit
...
+	metallb_check=$(kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' | grep customer-management | wc -l)
+fi
+
+# check SLS Networks data for BiCAN toggle
+curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l >> $TMPFILE ||
+    err_exit 30 "ERROR: Command failed (rc=$?): curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l" >> $TMPFILE
+sls_network_check=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l)
+
+# check spine switch motd for csm version
+ssh -o ConnectTimeout=1 -o BatchMode=yes-o StrictHostKeyChecking=no -o PasswordAuthentication=no admin@sw-spine-001 "" 2>&1 | grep CSM | awk '{ print $4 }' >> $TMPFILE ||
```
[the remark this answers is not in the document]
johren-hpe: This ssh doesn't work on Mellanox because it doesn't display the banner until you are logged in.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,46 @@ function err_exit
...
 # We really shouldn't be passing a password in plaintext on the command line, but as long as we are, at least
 # we won't also include it in the error message on failure
-if grep -q customer-management $TMPFILE ; then
-    echo "Running: canu validate network bgp --network all --password XXXXXXXX"
-    canu validate network bgp --network all --password {{.Env.SW_ADMIN_PASSWORD}} ||
-        err_exit 20 "canu validate network bgp --network all failed (rc=$?)"
-else
+# RFE: https://jira-pro.its.hpecorp.net:8443/browse/CASMNET-880
+
+if [ "$metallb_check" -eq "0" ] && [ "$sls_network_check" -eq "0" ] || [ "$switch_motd_check" == "1.0" ];then
```
johren-hpe, line 82: On an upgrade, SLS and switch are going to be at 1.2 right away but metallb is going to be at 1.0.
Therefore

( 1 && 0) || 0 == 0 so it will try to do all networks on an upgrade.

I think you want all ORs here.  If any of those checks indicate 1.0, then we need to only do NMN.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,46 @@ function err_exit
...
+    # csm-1.0 networking
     echo "Running: canu validate network bgp --network nmn --password XXXXXXXX"
     canu validate network bgp --network nmn --password {{.Env.SW_ADMIN_PASSWORD}} ||
-        err_exit 30 "canu validate network bgp --network nmn failed (rc=$?)"
+        err_exit 50 "canu validate network bgp --network nmn failed (rc=$?)"
+else
+    # csm-1.2+ networking
+    echo "Running: canu validate network bgp --network all --password XXXXXXXX"
+    canu validate network bgp --network all --password {{.Env.SW_ADMIN_PASSWORD}} ||
+        err_exit 60 "canu validate network bgp --network all failed (rc=$?)"
```
johren-hpe, line 91: With all of the above worked around, I'm still getting a failure logging in to the switches so this CASMINST-4284 is still broken.

```
ncn-m001:~/johren # ./test.sh
Running: canu validate network bgp --network nmn --password XXXXXXXX
Error connecting to switch 10.254.0.2, HTTPError.
/Error connecting to switch 10.254.0.3, HTTPError.


BGP Neighbors Established
--------------------------------------------------


Errors
--------------------------------------------------
10.254.0.2      - Connection Error
10.254.0.3      - Connection Error
PASS
```

johren-hpe (changes requested):

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,46 @@ function err_exit
...
+if [ "$metallb_cm" -gt "0" ];then
+	metallb_check="0"
+else
+    kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' | grep customer-management | wc -l >> $TMPFILE ||
+        err_exit 20 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' | grep customer-management | wc -l" > TMPFILE
+	metallb_check=$(kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' | grep customer-management | wc -l)
+fi
+
+# check SLS Networks data for BiCAN toggle
+curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l >> $TMPFILE ||
```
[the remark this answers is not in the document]
johren-hpe: Resolving this one because this shouldn't be an issue on a 1.2 upgrad or 1.0 install.

goss-testing/scripts/check_bgp_neighbors_established.sh
dle-hpe: revised testing items as suggested.  removed the error catch since we will have an error running this command on csm-1.0 system.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,46 @@ function err_exit
...
+kubectl -n metallb-system get cm metallb 2>&1 |grep -i error|wc -l > $TMPFILE ||
+    err_exit 10 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb 2>&1 |grep -i error|wc -l " >> $TMPFILE
+metallb_cm=$(kubectl -n metallb-system get cm metallb 2>&1 |grep -i error|wc -l)
+
+# Check for BiCAN configs in MetalLB configmap
+if [ "$metallb_cm" -gt "0" ];then
+	metallb_check="0"
+else
+    kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' | grep customer-management | wc -l >> $TMPFILE ||
+        err_exit 20 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' | grep customer-management | wc -l" > TMPFILE
```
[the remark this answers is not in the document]
dle-hpe: i am following the pattern that was pre-existing.  i can talk to Mitch about what the intentions were with the $TMPFILE

goss-testing/scripts/check_bgp_neighbors_established.sh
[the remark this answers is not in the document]
dle-hpe: removed motd checking as the casmnet group decided it was to brittle of a test to keep.

goss-testing/scripts/check_bgp_neighbors_established.sh
dle-hpe: Added logic to allow manual entry of password if goss env password is null or locally running test outside of goss.

goss-testing/scripts/check_bgp_neighbors_established.sh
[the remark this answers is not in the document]
johren-hpe: I believe previously the $TMPFILE was used in a later step.  But that is not the case anymore.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -31,6 +31,8 @@ set -o pipefail
 
 TMPFILE=/tmp/check_bgp_neighbors_established.$$.$RANDOM.tmp
 
+SWITCH_PASSWORD={{.Env.SW_ADMIN_PASSWORD}}
```
johren-hpe, line 34: This still won't work.  Bash doesn't know what to do with `{{.Env.SW_ADMIN_PASSWORD}}`

johren-hpe (changes requested):

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +51,47 @@ function err_exit
...
+else
+    metallb_check="1"
+fi
+
+# check SLS Networks data for BiCAN toggle
+echo "Checking for BiCAN toggle."  >> $TMPFILE
+curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l >> $TMPFILE
+sls_network_check=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l)
+
+if [ "$SWITCH_PASSWORD" == "{{.Env.SW_ADMIN_PASSWORD}}" ] || [ -z "$SWITCH_PASSWORD" ]; then
```
johren-hpe, line 73: What is the first conditional checking?

goss-testing/scripts/check_bgp_neighbors_established.sh
dle-hpe: ok fixed.  i had thought we had some sort of special setup

dzou-hpe (approved):

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,47 @@ function err_exit
...
+# If error checking for metallb configmap
+if [ "$metallb_cm" -gt "0" ];then
+	metallb_check="0"
+else
+    metallb_check="1"
+fi
+
+# check SLS Networks data for BiCAN toggle
+echo "Checking for BiCAN toggle."  >> $TMPFILE
+curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l >> $TMPFILE
```
mharding-hpe, line 68: This doesn't check to see if the curl or jq commands passed or not. I would replace this line and the next line with the following:
```bash
curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN |
    jq -r .ExtraProperties.SystemDefaultRoute |
    tee $TMPFILE ||
        err_exit 30 "Error writing to $TMPFILE or command pipeline failed with return code $?: curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN | jq -r .ExtraProperties.SystemDefaultRoute"
# Set sls_network_check to 0 if the string CHN or CAN appears in the output of the jq command
grep -e CHN -e CAN $TMPFILE
sls_network_check=$?
```

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,47 @@ function err_exit
     exit $rc
 }
 
-kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE ||
-    err_exit 10 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE"
+TOKEN=$(curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret=`kubectl get secrets admin-client-auth -o jsonpath='{.data.client-secret}' | base64 -d` https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token')
```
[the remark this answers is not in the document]
mharding-hpe: Actually, I missed that you have two commands embedded here. I would replace this line with the following:
```bash
SECRET=$(kubectl get secrets admin-client-auth -o jsonpath='{.data.client-secret}' | base64 -d) ||
    err_exit 10 "Command pipeline failed with return code $?: kubectl get secrets admin-client-auth -o jsonpath='{.data.client-secret}' | base64 -d"
# We omit the client secret from the error message, so it is not recorded in the log. 
# Ideally we would not be passing it to curl on the command line either.
TOKEN=$(curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret="$SECRET" https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token') ||
    err_exit 15 "Command pipeline failed with return code $?: curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret=XXXXXX https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token'"
```

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,47 @@ function err_exit
...
+# check SLS Networks data for BiCAN toggle
+echo "Checking for BiCAN toggle."  >> $TMPFILE
+curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l >> $TMPFILE
+sls_network_check=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN|jq -r .ExtraProperties.SystemDefaultRoute | grep -e CHN -e CAN |wc -l)
+
+if [ -z "$SW_ADMIN_PASSWORD" ]; then
+    echo "******************************************"
+    echo "******************************************"
+    echo "**** Enter SSH password of switches: ****"
+    read -t180 -sp ""  SWITCH_PASSWORD
```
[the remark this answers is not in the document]
dle-hpe: woops...

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,47 @@ function err_exit
     exit $rc
 }
 
-kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE ||
-    err_exit 10 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE"
+TOKEN=$(curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret=`kubectl get secrets admin-client-auth -o jsonpath='{.data.client-secret}' | base64 -d` https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token')
+
+# check if metalLB configmap exists
+echo "Checking for cofigmap called metallb." > $TMPFILE
+kubectl -n metallb-system get cm metallb 2>&1 |grep -i error|wc -l >> $TMPFILE
```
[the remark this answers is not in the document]
mharding-hpe: Although if you are grepping for "error" just as a way to see if the kubectl command failed, then this could be done much more straightforwardly:
```bash
# Set metallb_check to 0 if we cannot get the metallb configmap from Kubernetes. Set to 1 otherwise.
if kubectl -n metallb-system get cm metallb ; then
    metallb_check=1
else
    metallb_check-0
fi
```

I'm not sure if grep was chosen here because sometimes we expect the kubectl command to succeed but to still include the string "error" somewhere in its output. If that's the case, then my previous comment is more apporpriate.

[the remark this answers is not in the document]
dle-hpe: i need to check for the error as that particular configmap only exists in csm-1.2+

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,19 +49,47 @@ function err_exit
...
+
+# If error checking for metallb configmap
+if [ "$metallb_cm" -gt "0" ];then
+	metallb_check="0"
+else
+    metallb_check="1"
+fi
+
+# check SLS Networks data for BiCAN toggle
+echo "Checking for BiCAN toggle."  >> $TMPFILE
```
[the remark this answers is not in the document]
dle-hpe: i can remove both descriptive echo statements

goss-testing/scripts/check_bgp_neighbors_established.sh
dle-hpe: i need to check if exists and not stop on error since this is another scenario where this is a specific csm-1.2+ data element.

goss-testing/scripts/check_bgp_neighbors_established.sh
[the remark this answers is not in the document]
mharding-hpe: So you are saying there is a case where you would see the error string but the kubectl command would return a 0 exit code (meaning success)? Usually kubectl get exits non-0 if you try to retrieve something that doesn't exist, I think.

goss-testing/scripts/check_bgp_neighbors_established.sh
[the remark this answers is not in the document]
mharding-hpe: No no -- the echoes are good. Writing them to the TMPFILE is what I am confused about. You write them to the TMPFILE but then never do anything with it, and that file is deleted at the end of the script. You redirecting the output to that file, not teeing the output to it, so having the echoes this way means that their output is never seen.

goss-testing/scripts/check_bgp_neighbors_established.sh
mharding-hpe: Oh, so you are saying that there are some 1.0 systems where the curl command would succeed to that BICAN endpoint, but we just wouldn't see the CHN or CAN strings in the output from it?

mharding-hpe (changes requested):

goss-testing/scripts/check_bgp_neighbors_established.sh
mharding-hpe: Assuming you want to verify both that the BICAN endpoint works and that those strings are there, I would change this section to:
```bash
curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN |
    jq -r .ExtraProperties.SystemDefaultRoute |
    grep -e CHN -e CAN $TMPFILE
# Set sls_network_check to 1 if the BICAN endpoint responds and 
# has the CAN or CHN strings in its ExtraProperties.SystemDefaultRoute field. Otherwise set to 0
[[ $? -eq 0 ]] && sls_network_check=1 || sls_network_check=0
```

dle-hpe: another improvement!

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,20 +49,52 @@ function err_exit
...
-    err_exit 10 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE"
+SECRET=$(kubectl get secrets admin-client-auth -o jsonpath='{.data.client-secret}' | base64 -d) ||
+    err_exit 10 "Command pipeline failed with return code $?: kubectl get secrets admin-client-auth -o jsonpath='{.data.client-secret}' | base64 -d"
+# We omit the client secret from the error message, so it is not recorded in the log.
+# Ideally we would not be passing it to curl on the command line either.
+TOKEN=$(curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret="$SECRET" https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token') ||
+    err_exit 15 "Command pipeline failed with return code $?: curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret=XXXXXX https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token'"
+
+# check if metalLB configmap exists
+# Set metallb_check to 0 if we cannot get the metallb configmap from Kubernetes. Set to 1 otherwise.
```
mharding-hpe, line 60: The if statement is setting it to 0 if we *can* get the configmap from Kubernetes. So the if statement and this comment are in conflict.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,20 +49,52 @@ function err_exit
...
+    metallb_check=1
+fi
+
+# check SLS Networks data for BiCAN toggle
+curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN |
+    jq -r .ExtraProperties.SystemDefaultRoute |
+    grep -e CHN -e CAN
+# Set sls_network_check to 1 if the BICAN endpoint responds and
+# has the CAN or CHN strings in its ExtraProperties.SystemDefaultRoute field. Otherwise set to 0
+[[ $? -eq 0 ]] && sls_network_check=0 || sls_network_check=1
```
mharding-hpe, line 73: The comment and this line are saying opposite things.

mharding-hpe (changes requested):

mharding-hpe (approved):

.rpm_version
```diff
@@ -1 +1 @@
-1.12.19
+1.12.20
```
johren-hpe, line 1: This will need to be bumped to 21.

goss-testing/scripts/check_bgp_neighbors_established.sh
```diff
@@ -49,20 +49,52 @@ function err_exit
     exit $rc
 }
 
-kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE ||
-    err_exit 10 "ERROR: Command failed (rc=$?): kubectl -n metallb-system get cm metallb -o jsonpath='{.data.config}' > $TMPFILE"
```
johren-hpe, line 53: Now that TMPFILE is no longer being used, lines 32-39 above and any invocation of "cleanup" can be removed.

johren-hpe (changes requested):

goss-testing/scripts/check_bgp_neighbors_established.sh
dle-hpe: removed TMPFILE and cleanup function

.rpm_version
dle-hpe: version bump to 21

PR #258, opened 2022-03-25
casminst-4324: Remove goss-basecamp-json-ncns.yaml test from ncn-upgr…

rgross727:
### Summary and Scope
1.2.5 - CASMINST-4324: Remove goss-basecamp-json-ncns.yaml test from ncn-upgrade-preflight-tests suite

Remove goss-basecamp-json-ncns.yaml test from ncn-upgrade-preflight-tests suite.
Update the goss-basecamp-json-ncns.yaml test with pipefail check to prevent false positive result.
Update the goss-spire-bss-metadata-exist.yaml test with pipefail check to prevent false positive result.
Validate updated tests on wasp, 1.2.5-alpha.1

### Issues and Related PRs
CASMINST-4324

### Testing
Basecamp & Spire tests were validated on wasp with 1.2.5-alpha.1

Was a fresh Install tested? N - N/A
Was an Upgrade tested? N - N/A
Was a Downgrade tested? N - N/A
If schema changes were part of this change, how were those handled in your upgrade/downgrade testing? N/A

### Risks and Mitigations
Low

### Requires:
N/A


.rpm_version
```diff
@@ -1 +1 @@
-1.12.18
+1.12.19
```
johren-hpe, line 1: This is going to end up overwriting the 1.12.19 rpms from the 1.2 branch.   We need to figure out what we are going to do with versions if we are going to have a 1.2.5 branch.  I think we need further discussion on this, but for now we should at least bump the minor version so we are not overwriting 1.2 builds.

johren-hpe (changes requested):

.rpm_version
rgross727: Increase the 1.2.5 version to 1.13.0 now.

kburns-hpe (approved):

PR #273, opened 2022-04-13

mharding-hpe:
## Summary and Scope

CASMINST-4196 changed the goss-network-interfaces test so that the bond0.can check was skipped. That change was accidentally overwritten. This PR restores the change.

## Issues and Related PRs

* Resolves CASMINST-4446 (failure seen during today's install of wasp) and the original CASMINST-4196 ticket
* csm-1.2 PR https://github.com/Cray-HPE/csm-testing/pull/275
* csm-1.2.5 PR https://github.com/Cray-HPE/csm-testing/pull/274

## Testing

I tested this on wasp, where I hit the issue today, and verified that once I restored the change to the test, it worked fine.

## Risks and Mitigations

Low risk -- this change was already tested before, when it was originally made.

## Pull Request Checklist

- [ ] Copyrights updated
- [X] License file intact
- [X] Target branch correct
- [ ] CHANGELOG.md updated


goss-testing/tests/common/goss-network-interfaces.yaml
```diff
@@ -26,16 +26,18 @@
...
-            meta:
-                desc: Ensures VLAN was created for network interface '{{$interface}}'.
-                sev: 0
-            exec: |-
-                "{{$logrun}}" -l "{{$testlabel}}" \
-                    ip link show dev "{{$interface}}"
-            exit-status: 0
-            timeout: 10000
-            skip: false
+        {{if ne $interface "bond0.can0"}}
```
johren-hpe, line 29: This is not what we want.   Some systems will have the bond0.can0 interfaces and some will not depending on whether CAN or CHN was configured when the system was set up.   I think we may want to have a script here that checks the BICAN network configuration in SLS.

johren-hpe (changes requested):

goss-testing/tests/common/goss-network-interfaces.yaml
mharding-hpe: I'm fine with other options -- this is just the one that Jacob had implemented originally, in CASMINST-4196, and that I accidentally over-wrote in a PR. Right now the test fails on systems without bond0.can0, so unless we have a quick implementation of what you're suggesting, we may want to merge this PR if only as a stop-gap until we have a better solution.

If we do want to go that route, let me know, and I'll open another ticket to properly correct the test.

johren-hpe: I don't know if we will be doing many (or any) installs on wasp or hela in the next couple of days that would hit this issue again.   I feel like it will be better to have this tests the way it is to work on "CAN" systems (our more frequently installed systems) until we can get the fix that will address both "CAN" and "CHN" systems.

mharding-hpe: Okie dokie, makes sense to me. I'll close these PRs and update the ticket to reflect its new scope.

PR #274, opened 2022-04-13

mharding-hpe:
See the main branch PR for details:
https://github.com/Cray-HPE/csm-testing/pull/273


goss-testing/tests/common/goss-network-interfaces.yaml
```diff
@@ -26,16 +26,18 @@
...
-            meta:
-                desc: Ensures VLAN was created for network interface '{{$interface}}'.
-                sev: 0
-            exec: |-
-                "{{$logrun}}" -l "{{$testlabel}}" \
-                    ip link show dev "{{$interface}}"
-            exit-status: 0
-            timeout: 10000
-            skip: false
+        {{if ne $interface "bond0.can0"}}
```
johren-hpe, line 29: This is not what we want. Some systems will have the bond0.can0 interfaces and some will not depending on whether CAN or CHN was configured when the system was set up. I think we may want to have a script here that checks the BICAN network configuration in SLS.

johren-hpe (changes requested):

PR #275, opened 2022-04-13

mharding-hpe:
See the main branch PR for details:
https://github.com/Cray-HPE/csm-testing/pull/273


goss-testing/tests/common/goss-network-interfaces.yaml
```diff
@@ -26,16 +26,18 @@
...
-            meta:
-                desc: Ensures VLAN was created for network interface '{{$interface}}'.
-                sev: 0
-            exec: |-
-                "{{$logrun}}" -l "{{$testlabel}}" \
-                    ip link show dev "{{$interface}}"
-            exit-status: 0
-            timeout: 10000
-            skip: false
+        {{if ne $interface "bond0.can0"}}
```
johren-hpe, line 29: This is not what we want. Some systems will have the bond0.can0 interfaces and some will not depending on whether CAN or CHN was configured when the system was set up. I think we may want to have a script here that checks the BICAN network configuration in SLS.

johren-hpe (changes requested):

goss-testing/tests/common/goss-network-interfaces.yaml
johren-hpe: For example, wasp has BICAN set to CHN so it will not have bond0.can0 interfaces.
```
wasp-ncn-m001-pit:~ # cat /var/www/ephemeral/prep/wasp/sls_input_file.json | jq .Networks.BICAN
{
  "Name": "BICAN",
  "FullName": "SystemDefaultRoute points the network name of the default route",
  "IPRanges": [
    "0.0.0.0/0"
  ],
  "Type": "ethernet",
  "ExtraProperties": {
    "CIDR": "0.0.0.0/0",
    "VlanRange": [
      0
    ],
    "MTU": 9000,
    "Subnets": [],
    "SystemDefaultRoute": "CHN"
  }
}
```

But surtur has BICAN set to CAN so it does have bond0.can0 interfaces.
```
surtur-ncn-m001-pit:/var/www/ephemeral # cat /var/www/ephemeral/prep/surtur/sls_input_file.json | jq .Networks.BICAN
{
  "Name": "BICAN",
  "FullName": "SystemDefaultRoute points the network name of the default route",
  "IPRanges": [
    "0.0.0.0/0"
  ],
  "Type": "ethernet",
  "ExtraProperties": {
    "CIDR": "0.0.0.0/0",
    "VlanRange": [
      0
    ],
    "MTU": 9000,
    "Subnets": [],
    "SystemDefaultRoute": "CAN"
  }
}


ncn-w001:~ # ip addr show bond0.can0
6: bond0.can0@bond0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9000 qdisc noqueue state UP group default qlen 1000
    link/ether 14:02:ec:d9:7c:40 brd ff:ff:ff:ff:ff:ff
    inet 10.103.11.138/26 brd 10.103.11.191 scope global bond0.can0
       valid_lft forever preferred_lft forever
    inet6 fe80::1602:ecff:fed9:7c40/64 scope link
       valid_lft forever preferred_lft forever
```

PR #282, opened 2022-04-15
CASMINST-4457: Refine storage test

mharding-hpe:
## Summary and Scope

This PR contains 3 changes:
1. The pre-CSM deploy storage test previously used an unorthodox way to try and find a list of storage nodes, ran on the first one it found that worked, and then exited. However, some of its tests make sense to run on all of the storage nodes. Therefore, the `ncn-storage-tests` script was changed so that it used a more standard way to get the list of storage nodes (that is, it uses the same method the other automated test scripts do), and it runs the tests on all storage nodes. 
2. This will not significantly increase the runtime of the tests, as the only long-running test in the suite is modified so that it only runs on ncn-s001 (which is appropriate, since it runs with an option that checks all of the nodes anyway). 
3. While I was working on that test, I also modified it to run in verbose mode and to log its output.

## Issues and Related PRs

This is part of several PRs for CASMINST-4457. I broke them into different commits and PRs in order to simplify the review process.

## Testing

The testing for this PR needs to be done during an install, since it involves tests run from the PIT node. I plan to do that during my install on mug. But in the meantime I wanted to open this up for reviews.

THIS SHOULD NOT MERGE until I finish my testing.

## Risks and Mitigations

Medium risk. The changes to the automated test script are fairly large. On the other hand, they made it significantly simpler. After testing, I do not think there is a lot of risk.

## Pull Request Checklist

- [ ] Copyrights updated
- [X] License file intact
- [X] Target branch correct
- [ ] CHANGELOG.md updated

The version number will be added by another PR in the set.
I will check off the testing box above once that is done.

### DIFF
--- a/goss-testing/automated/ncn-storage-checks
+++ b/goss-testing/automated/ncn-storage-checks
@@ -27,91 +27,31 @@
 
 export GOSS_LOG_BASE_DIR=/opt/cray/tests/install/logs
 
-# Temporary file to store command output
-NCN_STORAGE_CHECKS_TMPFILE=/tmp/.ncn-storage-checks.tmp.$$.$RANDOM
-
 source "$GOSS_BASE/automated/run-ncn-tests.sh"
 
 echo $'\e[1;33m'Storage Node Automated Tests$'\e[0m'
 echo $'\e[1;33m'----------------------------$'\e[0m'
 
-# find the storage node names and query the server endpoint
-for FAKELOOP in 1 
-do
-    if ! command -v ceph &> /dev/null
-    then
-        echo "ceph command not found"
-    else
-        echo "ceph command found"
-        if ! ceph status -f json-pretty > ${NCN_STORAGE_CHECKS_TMPFILE} 2>&1
-        then
-            echo "Command failed: ceph status -f json-pretty"
-        elif ! mons=$(jq -r '.quorum_names[]' ${NCN_STORAGE_CHECKS_TMPFILE})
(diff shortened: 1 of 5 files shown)


goss-testing/automated/ncn-storage-checks
```diff
@@ -27,91 +27,31 @@
...
-
-# Remove the temporary file(s), if they exist
-rm -f ${NCN_STORAGE_CHECKS_TMPFILE} ${NCN_STORAGE_CHECKS_TMPFILE}.2 >& /dev/null
-
-(( counter=0 ))
-
-for node in $mons; do
-    nodename=$(echo $node|cut -d":" -f1)
-    echo "$counter: Checking $nodename"
-    if [[ $(nc -z -w 10 $node 22) ]] || [[ $counter -lt 3 ]]
```
kimjensen-hpe, line 104: Do we know that the netcat command is not needed anymore to verify that the storage node is alive and responding?

kimjensen-hpe (approved):

johren-hpe (approved):

goss-testing/automated/ncn-storage-checks
mharding-hpe: Good question. I think we're okay here - we basically cover this implicitly with this PR, because we try to run the goss tests on all storage nodes. For any which aren't up (or not on the network, anyway), we'll get a failure in trying to run the goss tests.

PR #304, opened 2022-05-05
CASMINST-4576: Validate that ceph -s works where and when it ought to

mharding-hpe:
## Summary and Scope

We ran into a problem during an install on gamora that could have been detected earlier than it was. ncn-m002 was not fully configured and this would have been caught had we been validating that "ceph -s" worked on there. Currently the only tests we have which do that are run only on storage nodes.

This PR moves the "ceph -s" validation into its own test. The test has logic included so that it will not try to run "ceph -s" on storage nodes beyond ncn-s003.

This PR adds the new test to the following suites:
* **ncn-storage-tests** (this runs on storage nodes after NCN deploy)
* **ncn-kubernetes-tests-master** (this runs on master nodes after NCN deploy AND during CSM health validation)
* **ncn-healthcheck-storage** (this runs on storage nodes during CSM health validation)

While I was at it, I sorted the tests being run in those 3 suites alphabetically, just to make it easier to determine if a particular test is included (since they tests in the suite are all executed simultaneously anyway, so the ordering doesn't matter from that perspective).

Finally, I fixed an oversight in the **ncn-kubernetes-checks** script that was causing most tests to not be run on ncn-m001 even after PIT redeploy. As part of this, I also moved some tests that were being called individually from that script and put them into test suites, and called those instead. The two new suites are:
* **common-kubernetes-tests-cluster** (this runs locally on the PIT node when the **ncn-kubernetes-checks** script is run there)
* **ncn-kubernetes-tests-cluster** (this runs locally on whatever NCN the **ncn-kubernetes-checks** script is run from, but not if it is the PIT node. This suite calls the previous suite, so it is a superset of that suite).

These two suites are where we should put any tests which only need to be run once on the Kubernetes cluster, rather than on every single worker and master node. I will leave the task of moving tests over into the suites for later PRs, since this one has grown enough as it is.

## Testing

I ran this on mug (csm-1.2).

## Risks and Mitigations

Low risk.

## Pull Request Checklist

- [X] License file intact
- [X] Target branch correct


goss-testing/tests/ncn/goss-ceph-status.yaml
```diff
@@ -0,0 +1,50 @@
...
+    {{if $thisnode | regexMatch "^ncn-(m|s00[1-3])" }}
+        {{ $testlabel := "ceph_s_runs" }}
+        {{$testlabel}}:
+            title: ceph -s
+            meta:
+                desc: Check that ceph -s successfully executes
+                sev: 0
+            exec: |-
+                "{{$logrun}}" -l "{{$testlabel}}" \
+                    /usr/bin/ceph -s
```
cdelatte-hpe, line 47:
```suggestion
                    /usr/bin/ceph health
```

cdelatte-hpe: it is a little cleaner and only gives the HEALTH_OK/WARN/.....

cdelatte-hpe (changes requested):
I minor thing to make a little cleaner

rustydb (approved):

goss-testing/tests/ncn/goss-ceph-status.yaml
mharding-hpe: Since the output is only seen if the test failure needs to be debugged, it seems like it would be more helpful to have it show the additional information. Otherwise wouldn't they still need to run it again to know why it failed?

mharding-hpe: I already merged the 1.2 PR before I saw your comment on this one. I'm going to go ahead and merge this one as-is, to keep it common, but we can use an additional PR to change them both to use the more terse option, if desired.

PR #307, opened 2022-05-09
CASMINST-4446: Fix goss-network-interfaces test to handle CHN and CAN configurations

johren-hpe:
## Summary and Scope

This test will always check to see if the bond0.can0 interface exists on the worker and master nodes.   However, this interface not exist if the system was configured with CHN instead of CAN.   This changes the test into a script that checks the BICAN network configuration in SLS to determine if it is CAN or CHN.   It will only check for the bond0.can0 interface if it is configured for CAN.

## Issues and Related PRs

* Resolves CASMINST-4446

### Tested on:

  * `surtur` and `wasp`

### Test description:

I ran this goss test on wasp where it is configured for CHN and previous failed.  It now passes there.
I also ran this goss test on surtur where it was configured for CAN to make sure it still passes there.

Wasp:
```
ncn-m001:~ # GOSS_BASE=/opt/cray/tests/install/ncn goss -g /opt/cray/tests/install/ncn/tests/goss-network-interfaces.yaml --vars=/opt/cray/tests/install/ncn/vars/variables-ncn.yaml validate
....

Total Duration: 0.283s
```

I also ran through several injected negative test cases (not able to get secret, not able to pull token, etc.) to test the error checking.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [ ] CHANGELOG.md updated


goss-testing/scripts/check_network_interface.sh
```diff
@@ -0,0 +1,72 @@
...
+{
+    local rc
+    rc=$1
+    shift
+    echo "ERROR: $*" 1>&2
+    echo "FAIL"
+    exit $rc
+}
+
+TESTIF=$1
```
mharding-hpe, line 42: I don't know why, but I always like to validate the inputs to scripts. If nothing else it can detect if something upstream has gone haywire and is passing in weird stuff.

In terms of echoing the argument, the test logger I wrote will log this itself, but I like to print the arguments to scripts. This is mainly in case someone calls it manually, just because every once in a while someone uses a shell loop to run a script and end up passing in an unexpected value. So I like to show them what they called it with.

```suggestion
[ $# -eq 1 ] || err_exit 5 "Test requires exactly 1 argument but received $#: $*"
[ -n "$1" ] || err_exit 7 "Argument to test may not be blank"

TESTIF=$1

echo "Test interface: $TESTIF"
```

goss-testing/scripts/check_network_interface.sh
```diff
@@ -0,0 +1,72 @@
...
+if [ -z "$SECRET" ]; then
+    err_exit 12 "Failed to retrieve admin-client-auth secret"
+fi
+
+TOKEN=$(curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret="$SECRET" https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token') ||
+    err_exit 15 "Command pipeline failed with return code $?: curl -s -k -S -d grant_type=client_credentials -d client_id=admin-client -d client_secret=XXXXXX https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token | jq -r '.access_token'"
+if [ -z "$TOKEN" ] || [ "$TOKEN" == null ]; then
+    err_exit 17 "Failed to retrieve token from https://api-gw-service-nmn.local/keycloak/realms/shasta/protocol/openid-connect/token"
+fi
+
```
mharding-hpe, line 55:
```suggestion
```

goss-testing/scripts/check_network_interface.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# check SLS Networks data for BiCAN toggle
+USER_NETWORK=$(curl -s -k -H "Authorization: Bearer ${TOKEN}" https://api-gw-service-nmn.local/apis/sls/v1/networks/BICAN | jq -r .ExtraProperties.SystemDefaultRoute | tr '[:upper:]' '[:lower:]')
+if [ -z "$USER_NETWORK" ] || [ "$USER_NETWORK" == null ]; then
```
mharding-hpe, line 59: I suspect the other two checks are sufficient to cover this anyway, but maybe add a return code check here.
```suggestion
if [ $? -ne 0 ] || [ -z "$USER_NETWORK" ] || [ "$USER_NETWORK" == null ]; then
```

goss-testing/scripts/check_network_interface.sh
```diff
@@ -0,0 +1,72 @@
...
+    err_exit 18 "Failed to retrieve user network from SLS"
+fi
+
+if [[ "$TESTIF" =~ .*can0 ]]; then
```
mharding-hpe, line 63:
```suggestion
echo "USER_NETWORK = $USER_NETWORK"

if [[ "$TESTIF" =~ .*can0 ]]; then
```

goss-testing/scripts/check_network_interface.sh
```diff
@@ -0,0 +1,72 @@
...
+    if [[ "$TESTIF" =~ .*"$USER_NETWORK".* ]]; then
+        ip link show dev $TESTIF || err_exit 20 "Command pipeline failed with return code $?: ip link show dev $TESTIF"
+    fi
```
mharding-hpe, line 66: This would be easy enough to deduce by the absence of the output from the ip link command, 
```suggestion
    else
        echo "Skipping test because CAN is not the user network"
    fi
```

mharding-hpe (approved):

goss-testing/scripts/check_network_interface.sh
johren-hpe: Will address in a later PR

goss-testing/scripts/check_network_interface.sh
johren-hpe: Will address in a later PR

goss-testing/scripts/check_network_interface.sh
johren-hpe: Will address in a later PR

goss-testing/scripts/check_network_interface.sh
johren-hpe: Will address in a later PR

PR #345, opened 2022-07-21
CASMINST-5025: Create more shared functions for automated scripts; prove methods for determing NCN lists

mharding-hpe:
### TL;DR

* Create common function to determine if script is running on the PIT node; update other scripts to call it.
* Create common function to generate list of NCNs; update other scripts to call it.
* Update goss-k8s-nodes-age-valid and goss-k8s-nodes-joined-cluster tests so they realize ncn-m001 is not in the cluster if they are running on the PIT node.
* Update value of kubectl variable for the PIT node
* Create common goss variable file for variables with same value between PIT and NCN; Modify RPM builder so that it generates NCN/PIT variable files by combining common file with PIT/NCN file.
* Moved a couple of common functions into the common function library.
* General linting and cleanup to make the scripts more consistent.

### Details

This ticket began just as a fix to a problem being seen on the surtur install, where the dnsmasq statics file had been deleted from the PIT node at some point, which caused the csm-testing scripts to assume that it was a normal NCN. This caused other problems.

While there is no surefire way to determine the node identity if someone may be going around changing things in the environment (like deleting the dnsmasq file), it seems better to key off something like the pit-release file regardless. So this PR was originally going to make that change. But then I realized that there was code all across the automated scripts that was deciding whether or not the particular script was running on the PIT node. So I used this PR to change all of those to instead use a shared function for this, so we don't play whack-a-mole when individual scripts mess up.

In the process of doing that, I also noticed some other code that could be made common, in order to make the scripts more consistent and less error prone. And I realized that if we were assuming that the dnsmasq file may not be there, then our code for making a list of NCNs would need to change. And this is code which existed in different forms in several places. So another common function I created was one to get the list of NCNs on the system (either all of them, or particular types, with an option to exclude the PIT). 

In the process of testing this, I discovered that a couple of Kubernetes checks were being made for ncn-m001 even though the test was running from the booted PIT node. So I modified those tests to omit ncn-m001 when executing on the PIT node.

While doing all of that, I tried to do other minor cleanup, linting, etc, to make the scripts more consistent in appearance and convention.

And finally, I noticed on surtur today that path for kubectl on the PIT node had changed, which tripped some of the goss tests up, so I am correcting that here as well. When doing that, I noticed that we had a fair number of variables which were in common from NCNs and LiveCDs. So I created a common variables file with the shared values, and then added a little code in the RPM builder to have it create the installed variable files based on a merger of the PIT/NCN variable file and the common variable file.

## Issues and Related PRs

https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5025

## Testing

I tested the changes on surtur by running all of the automated scripts on both the PIT node and on other NCNs. In addition, I made sure that the updated goss service script file was correctly populating the variable file and starting the servers.

## Risks and Mitigations

Moderate risk in the sense that a lot of code was touched. But the result is a much more maintainable situation than we had before, so even if problems are found that eluded my testing on surtur, it should be much easier to correct them.

## Pull Request Checklist

- [X] License file intact
- [X] Target branch correct

### DIFF
--- a/csm-testing.spec
+++ b/csm-testing.spec
@@ -58,12 +58,17 @@ install -d -m 755 %{buildroot}%{ncn}/vars
 install -m 755 goss-testing/automated/*         %{buildroot}%{livecd}/automated
 install -m 755 goss-testing/tests/livecd/*.yaml %{buildroot}%{livecd}/tests
 install -m 755 goss-testing/tests/common/*.yaml %{buildroot}%{livecd}/tests
-install -m 755 goss-testing/vars/*.yaml         %{buildroot}%{livecd}/vars
 install -m 755 build-testing/*                  %{buildroot}%{livecd}/build-testing
 install -m 755 goss-testing/automated/*         %{buildroot}%{ncn}/automated
 install -m 755 goss-testing/tests/ncn/*.yaml    %{buildroot}%{ncn}/tests
 install -m 755 goss-testing/tests/common/*.yaml %{buildroot}%{ncn}/tests
-install -m 755 goss-testing/vars/*.yaml         %{buildroot}%{ncn}/vars
+# Install variables files
+install -m 755 goss-testing/vars/vars-packages.yaml         %{buildroot}%{livecd}/vars
+install -T -m 755 goss-testing/vars/variables-common.yaml %{buildroot}%{livecd}/vars/variables-livecd.yaml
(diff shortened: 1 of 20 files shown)


rustydb (approved):

rustydb:
Good changes overall, and these look to be tested so I approved.

mharding-hpe:
The RPMs for this are currently installed on surtur (all NCNs and PIT), if anyone wants to check anything in particular.

kimjensen-hpe (approved):

PR #349, opened 2022-08-01
CASMINST-5159: Speed up Goss test result reporting

mharding-hpe:
## Summary and Scope

A few of the Goss automated scripts use Bash scripting and a lot of jq commands to parse the results and display them. This is very slow and can take longer than the tests themselves, more than doubling the time of the script execution.

This PR creates a Python script which replaces that code. The script makes a GET request to a Goss endpoint, or reads Goss JSON results from stdin (or from a file). Either way, it then parses the JSON results and presents them in the same way that the current scripts do. Just much faster (see testing section for details).

[CASMINST-5159](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5159) is phase 1 of a multi-part plan. The next stage is [CASMINST-5161](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5161), to modify our scripts which make calls to multiple goss endpoints so that they can present a grand total result summary at the end of the output, thus removing the need for our current "grep for the total lines" stuff in our docs. The goal here would also be to only output failures (if any) and a summary of execution -- not the current page after page of test successes. That output would instead be sent to a log file.

The final phase is [CASMINST-5162](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5162), to, if our tests don't have conflicts that prevent it, hit  the test endpoints in parallel on the nodes.

But for this PR, it just replaces the JSON parsing with something much faster.

## Testing

Three files contained this parsing code:
livecd-preflight-checks
livecd-provisioning-checks
run-ncn-tests.sh

I tested the two livecd scripts on the surtur PIT node. 

I tested the run-ncn-tests.sh changes by calling the ncn-healthcheck script on rocket, because it in turn calls the three ncn-healthcheck-master, -storage, and -worker scripts, and each of those call the modified function in run-ncn-tests.sh. 

I ran the scripts both with and without this PR in place. The results were the same for both in terms of the tests being executed and the output reported (as in, no problems were introduced).  However, the speeds were improved:

Script | Time without PR | Time with PR 
------ | ------------------ | ---------------
livecd-preflight-checks | 54.364s | 17.516s
livecd-provisioning-checks | 3.286s | 0.398s
ncn-healthcheck | 5m27.408s | 3m10.443s

In addition, I made sure that I tested cases where there were test failures, to ensure that they were properly handled.

## Risks and Mitigations

The only change is to the logic for parsing the test output. Low risk.

## Pull Request Checklist

- [X] License file intact
- [X] Target branch correct

### DIFF
--- a/goss-testing/automated/livecd-preflight-checks
+++ b/goss-testing/automated/livecd-preflight-checks
@@ -29,7 +29,7 @@ set -e
 
 # GOSS_BASE should already be set on the PIT node, but just in case we
 # will set it here
-export GOSS_BASE=${GOSS_BASE-"/opt/cray/tests/install/livecd"}
+export GOSS_BASE=${GOSS_BASE:-"/opt/cray/tests/install/livecd"}
 
 source "${GOSS_BASE}/automated/run-ncn-tests.sh"
 
--- a/goss-testing/automated/livecd-preflight-checks
+++ b/goss-testing/automated/livecd-preflight-checks
@@ -29,9 +29,7 @@ set -e
 
 # GOSS_BASE should already be set on the PIT node, but just in case we
 # will set it here
-[[ -z ${GOSS_BASE} ]] && export GOSS_BASE="/opt/cray/tests/install/livecd"
-
-export GOSS_LOG_BASE_DIR=/opt/cray/tests/install/logs
+export GOSS_BASE=${GOSS_BASE-"/opt/cray/tests/install/livecd"}
 
 source "${GOSS_BASE}/automated/run-ncn-tests.sh"
 
@@ -54,57 +52,9 @@ echo $'\e[1;33m'Running LiveCD preflight checks \(may take a few minutes to comp
 
 # Run the LiveCD preflight Goss test suite
 export GOSS_VARS=${tmpvars}
-set +e
-results=$(/usr/bin/goss validate -f json)
-set -e
 
-if ! `echo ${results} | jq -e > /dev/null 2>&1`; then
-  err_exit $'\e[1;31m'Output not valid JSON$'\e[0m'
-fi
+/usr/bin/goss validate -f json | "${GOSS_BASE}/automated/print_goss_json_results" --input - ||
+	err_exit $'\e[1;31m'"There were test errors"$'\e[0m'
 
-echo ${results} | jq -c '.results | sort_by(.result) | .[]' | while read -r test; do
-  result=$(echo ${test} | jq -r '.result')
-
-  if [[ -z ${result} ]]; then
-    continue
-  elif [[ ${result} == 0 ]]; then
-    result=PASS
-    echo $'\e[1;32m'
-  else
-    result=FAIL
-    echo $'\e[1;31m'
-  fi
-
-  title=$(echo ${test} | jq -r '.title')
-  description=$(echo ${test} | jq -r '.meta.desc')
-  severity=$(echo ${test} | jq -r '.meta.sev')
-  summary=$(echo ${test} | jq -r '."summary-line"')
-  time=$(echo ${test} | jq -r '.duration')
-  time=$(echo "scale=8; ${time}/1000000000" | bc | awk '{printf "%.8f\n", $0}')
-
-  echo "Result: ${result}"
-  echo "Test Name: ${title}"
-  echo "Description: ${description}"
-  echo "Severity: ${severity}"
-  echo "Test Summary: ${summary}"
-  echo "Execution Time: ${time} seconds"
-  echo "Node: ncn-m001"
-done
-
-echo $'\e[0m'
-
-total=$(echo ${results} | jq -r '.summary."test-count"')
-failed=$(echo ${results} | jq -r '.summary."failed-count"')
-time=$(echo ${results} | jq -r '.summary."total-duration"')
-time=$(echo "scale=4; ${time}/1000000000" | bc | awk '{printf "%.4f\n", $0}')
-
-echo "Total Tests: ${total}, Total Passed: $((total-failed)), Total Failed: ${failed}, Total Execution Time: ${time} seconds"
-echo
-
-if [[ ${total} -eq 0 ]]; then
-    err_exit "No tests were run!"
-elif [[ ${failed} -ne 0 ]]; then
-    err_exit "Not all tests passed"
-fi
 echo "All tests passed!"
 exit 0
--- a/goss-testing/automated/print_goss_json_results
+++ b/goss-testing/automated/print_goss_json_results
@@ -1,0 +1,218 @@
+#!/usr/bin/env python3
+#
+# MIT License
+#
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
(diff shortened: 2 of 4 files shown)


rustydb (approved):

rgross727 (approved):

goss-testing/automated/livecd-preflight-checks
```diff
@@ -29,9 +29,7 @@ set -e
 
 # GOSS_BASE should already be set on the PIT node, but just in case we
 # will set it here
-[[ -z ${GOSS_BASE} ]] && export GOSS_BASE="/opt/cray/tests/install/livecd"
-
-export GOSS_LOG_BASE_DIR=/opt/cray/tests/install/logs
+export GOSS_BASE=${GOSS_BASE-"/opt/cray/tests/install/livecd"}
```
mitchty, line 32: If GOSS_BASE is null do we really want it to still be null with this assignment? Otherwise I think you want parameter expansion of :- not - which if set and null substitutes word, ref section 2.6.2 of https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html

(true of all these scripts not going to mark on each)

General gist is if you're wanting to assign if the var is set otherwise use this if it is null or not set you want :-, - is pretty rare usage wise especially for an export.

goss-testing/automated/print_goss_json_results
```diff
@@ -0,0 +1,218 @@
...
+def debug_log_values(**kwargs):
+    logger.debug(" ".join([ f"{k}={v}" for (k, v) in kwargs.items() ]))
+
+class ScriptException(Exception):
+    pass
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Print JSON-format Goss test results with pretty colors")
+    parser.add_argument("--input", default="-", metavar="input_source", help="Source file or URL for test results. Defaults to stdin.")
+    parser.add_argument("--node_name", help="Node name to include in result stanzas ")
```
mitchty, line 94: Nit: extra space in the help text.

goss-testing/automated/print_goss_json_results
```diff
@@ -0,0 +1,218 @@
...
+        try:
+            with open(input_file, "rt") as infile:
+                input = infile.read()
+        except OSError:
+            logger.error(traceback.format_exc())
+            traceback.print_exc()
+            raise ScriptException(f"Problem reading input file {input_file}")
+    try:
+        return json.loads(input)
+    except json.decoder.JSONDecodeError:
```
mitchty, line 114: Might want a general except here too in case you get something like an EPERM or non OSErrror/jsondecode error(s) if only to clean up the stack trace dumped to users of the script.

goss-testing/automated/run-ncn-tests.sh
```diff
@@ -45,6 +45,22 @@ function is_pit_node {
...
+GOSS_INSTALL_BASE_DIR="/opt/cray/tests/install"
+
+if [[ -z ${GOSS_BASE} ]]; then
+    if is_pit_node ; then
+        GOSS_BASE="${GOSS_INSTALL_BASE_DIR}/livecd"
+    else
+        GOSS_BASE="${GOSS_INSTALL_BASE_DIR}/ncn"
+    fi
+fi
+export GOSS_BASE
```
mitchty, line 60: Shouldn't this be up a line? if the -z doesn't pass this could set GOSS_BASE to null and export it into the env.

goss-testing/automated/livecd-preflight-checks
mharding-hpe: Good catch!

goss-testing/automated/run-ncn-tests.sh
mharding-hpe: If the -z doesn't pass, that would mean it was already set. In that case, I believe this export command would export it with its current value. Is that not correct?

mitchty: True, though it could be null and that still passes it through though for that it might be better to do export GOSS_BASE=${GOSS_BASE:?} so if its set and null or unset it exits so that things can't trickle down further. All depends on if a set and null value should(n't) get passed into the env of anything fork()/exec()'d later.

mharding-hpe: Do you mean null as it just whitespace? I think if it's null as in no value at all, -z would fail, wouldn't it?

mharding-hpe:
```
root@jolt1-ncn-m001-pit 2022-08-01 20:34:54 /var/log/conman # D=""
root@jolt1-ncn-m001-pit 2022-08-01 20:39:29 /var/log/conman # [[ -z ${D} ]]
root@jolt1-ncn-m001-pit 2022-08-01 20:39:36 /var/log/conman # echo $?
0
```

goss-testing/automated/livecd-preflight-checks
mharding-hpe: I made changes to address this (in all of the files). Please mark this resolved if you think it covers it.

goss-testing/automated/print_goss_json_results
mharding-hpe: I made changes to address this. Please mark this resolved if you agree (I'd do it myself, but I like to make sure the person who commented is in agreement).

goss-testing/automated/print_goss_json_results
mharding-hpe: I made changes to address this. Please mark this resolved if you agree (I'd do it myself, but I like to make sure the person who commented is in agreement).

goss-testing/automated/run-ncn-tests.sh
mitchty: Not exactly (bash but all posix sh should behave the same):

```$ (export FOO=; [ -z $FOO ] && echo set and null || echo not ok )
set and null
$ (export FOO=blah; [ -z $FOO ] && echo set and not null || echo not ok )
not ok``

Comment (mharding-hpe) [reply goss-testing/automated/run-ncn-tests.sh:60]: I'm confused here. When you have it set and null, the -z is returning return code 0, right? Which means that the if stanza in question here would have set the variable, wouldn't it?

Comment (mitchty) [reply goss-testing/automated/run-ncn-tests.sh:60]: Sorry forgot the third possibility:

```
$ ([ -z $FOO ] && echo not set || echo not ok)
not set
```

Comment (mharding-hpe) [reply goss-testing/automated/run-ncn-tests.sh:60]: This is the code in question:

```
if [[ -z ${GOSS_BASE} ]]; then
    if is_pit_node ; then
        GOSS_BASE="${GOSS_INSTALL_BASE_DIR}/livecd"
    else
        GOSS_BASE="${GOSS_INSTALL_BASE_DIR}/ncn"
    fi
fi
export GOSS_BASE
```

If GOSS_BASE starts out set and null, or not set at all, either way it will get set inside the if clause, won't it?

mitchty: Although thinking about it you're right, ignore this chain, my brains becoming a pumpkin.

mharding-hpe: I'm trying to figure out what starting situation would result in GOSS_BASE being exported with a null value

mharding-hpe: Whew. Okay. I thought I was losing my mind. :) I mean, moreso than usual.

PR #362, opened 2022-08-11
MTL-1847 / simplified boot param validation

jpdavis-prof:
## Summary and Scope

modified boot param validation to only check for required boot params.

## Issues and Related PRs

* Resolves: https://jira-pro.its.hpecorp.net:8443/browse/MTL-1646
* Resolves: https://jira-pro.its.hpecorp.net:8443/browse/MTL-1847

### Tested on:

* drax
* fanta
* mug
* redull pit
* locally

### Test description:

Executed on the above nodes and saw no issues with reading and verifying Basecamp / BSS data.

## Risks and Mitigations
The script only reads and verifies data in BSS / Basecamp, so risk is none.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/goss-testing/scripts/data-validator.py
+++ b/goss-testing/scripts/data-validator.py
@@ -227,16 +227,14 @@ def boot_params(data):
 	"console=tty0",
 	"console=ttyS0,115200",
 	"crashkernel=[0-9]*M",
-	"ds=nocloud-net;s=http://",
+	"ds=",
 	"hostname=",
 	"ifname=mgmt0",
 	"ifname=mgmt1",
 	"initrd=initrd.img.xz",
 	"iommu=pt",
 	"log_buf_len=1",
-	"metal.no-wipe=",
-	"metal.no-wipe=1",
-	"metal.server=http://",
+	"metal.server=",
 	"pcie_ports=native",
 	"psi=1",
 	"rd.auto=1",
@@ -246,7 +244,7 @@ def boot_params(data):
 	"rd.live.overlay.thin=1",
 	"rd.live.overlay=LABEL=ROOTRAID",
 	"rd.live.ram=0",
-	"rd.live.squashimg=filesystem.squashfs",
+	"rd.live.squashimg=",
 	"rd.luks.crypttab=0",
 	"rd.lvm.conf=0",
 	"rd.lvm=1",
@@ -295,7 +293,7 @@ def boot_params(data):
         if not re.search(param, blob['params']):
           print_err(f"{hostname}: {param} not found in boot params")
           err = 1
-      
+
   if err == 1:
     return err
 
--- a/goss-testing/scripts/data-validator.py
+++ b/goss-testing/scripts/data-validator.py
@@ -277,15 +277,15 @@ def boot_params(data):
     if "params" in blob and blob['cloud-init']['user-data'] is not None:
       hostname = blob['cloud-init']['user-data']['hostname']
 
-      if re.search("ncn-w00", hostname):
+      if re.search("ncn-w", hostname):
         mod_params = ncn_params
         mod_params = mod_params + worker_params
 
-      elif re.search("ncn-s00", hostname):
+      elif re.search("ncn-s", hostname):
         mod_params = ncn_params
         mod_params = mod_params + storage_params
 
-      elif re.search("ncn-m00", hostname):
+      elif re.search("ncn-m", hostname):
         mod_params = ncn_params
         mod_params = mod_params + management_params
 
--- a/goss-testing/scripts/data-validator.py
+++ b/goss-testing/scripts/data-validator.py
@@ -277,15 +277,15 @@ def boot_params(data):
     if "params" in blob and blob['cloud-init']['user-data'] is not None:
       hostname = blob['cloud-init']['user-data']['hostname']
 
-      if re.search("ncn-w0", hostname):
+      if re.search("ncn-w00", hostname):
         mod_params = ncn_params
         mod_params = mod_params + worker_params
 
-      elif re.search("ncn-s0", hostname):
+      elif re.search("ncn-s00", hostname):
         mod_params = ncn_params
         mod_params = mod_params + storage_params
 
-      elif re.search("ncn-m0", hostname):
+      elif re.search("ncn-m00", hostname):
         mod_params = ncn_params
         mod_params = mod_params + management_params
 
--- a/goss-testing/scripts/data-validator.py
+++ b/goss-testing/scripts/data-validator.py
@@ -38,6 +38,7 @@ def print_err(*a):
 def get_data():
   """Get data from data.json or BSS.
   """
+
   hostname = socket.gethostname()
 
   if os.path.isfile("/etc/pit-release"):
@@ -66,6 +67,7 @@ def get_data():
 def user_data(data):
   """Returns the 'user-data' blobs from data.json or BSS.
   """
+
   filtered_data = []
 
   # if we're using BSS data
@@ -84,6 +86,9 @@ def user_data(data):
 
 
 def drill_down_data(data, desired_keys):
+  """Returns target data.
+  """
+
   for i in range(0, ( len(desired_keys) + 1), 1):
     if i == len(desired_keys):
       return saved_data
@@ -106,6 +111,7 @@ def drill_down_data(data, desired_keys):
 def are_valid_ip_masks(data, desired_keys):
   """Checks that IP/mask or CIDR is valid.
   """
+
   filtered_data = user_data(data)
   err = 0
 
@@ -117,6 +123,7 @@ def are_valid_ip_masks(data, desired_keys):
     if not child_key:
       print_err(f"ERR: {desired_keys} is not defined for: {instance_hostname}")
       err = 1
+
     else:
       for value in child_key:
         try:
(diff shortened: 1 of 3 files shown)


goss-testing/scripts/data-validator.py
```diff
@@ -192,103 +206,98 @@ def are_hosts_sane(data, desired_keys):
...
+	"pcie_ports=native",
+	"psi=1",
+	"rd.auto=1",
+	"rd.bootif=0",
+	"rd.dm=0",
+	"rd.live.overlay.overlayfs=1",
+	"rd.live.overlay.thin=1",
+	"rd.live.overlay=LABEL=ROOTRAID",
+	"rd.live.ram=0",
+	"rd.live.squashimg=filesystem.squashfs",
```
rustydb, line 249: This should only verify the prefix `rd.live.squashimg`, the filename could change in general. I know it'll change for the upcoming pre-signed URL work too.

This'll eventually be `rd.live.squashimg=rootfs` with the pre-signed URL work, but in general it might be better to just check for `rd.live.squashimg` or make this one optional.

goss-testing/scripts/data-validator.py
```diff
@@ -192,103 +206,98 @@ def are_hosts_sane(data, desired_keys):
...
+	"crashkernel=[0-9]*M",
+	"ds=nocloud-net;s=http://",
+	"hostname=",
+	"ifname=mgmt0",
+	"ifname=mgmt1",
+	"initrd=initrd.img.xz",
+	"iommu=pt",
+	"log_buf_len=1",
+	"metal.no-wipe=",
+	"metal.no-wipe=1",
```
rustydb, line 238: Why is this listed twice?

I'd actually advocate for not checking for `metal.no-wipe`. There's too many permutations:
- fresh install will have `metal.no-wipe=0`
- a finished install will have `metal.no-wipe=1`
- a disk boot always has `metal.no-wipe=1`
- a rebuild of any NCN uses `metal.no-wipe=0`
- an upgrade of a k8s-master and k8s-worker uses `metal.no-wipe=0`
- an upgrade of a storage-ceph node uses `metal.no-wipe=1`

Maybe ignoring this one would be easier?

goss-testing/scripts/data-validator.py
```diff
@@ -192,103 +206,98 @@ def are_hosts_sane(data, desired_keys):
...
+	"metal.server=http://",
```
rustydb, line 239: This will vary between `metal.server=http://` and `metal.server=s3://`

- `http` is going to be on the cmdline
- `s3` will be in BSS, but will show up as `http` on the cmdline because BSS swaps it out before plugging the cmdline into the node (so it'll show up as `s3` when comparing to BSS, despite looking like `http` on the cmdline)

rustydb: Eventually this could even vary between `https` and `s3`.

Checking for `s3` in BSS is a good test, but idk if this can be tested against basecamp.

Also this param should *not* exist in a disk cmdline, it should *only* exist in a netbooting cmdline.

goss-testing/scripts/data-validator.py
```diff
@@ -192,103 +206,98 @@ def are_hosts_sane(data, desired_keys):
...
-  macreg = re.compile(r':([0-9a-fA-F]){2}:([0-9a-fA-F]){2}:([0-9a-fA-F]){2}:([0-9a-fA-F]){2}:([0-9a-fA-F]){2}:([0-9a-fA-F]){2}')
-  hostreg = re.compile(r'hostname=[a-zA-Z]*\-[a-zA-Z0-9]*')
-  httpreg = re.compile(r's=http://.*/')
+  err = 0
+  ncn_params = [
+	"biosdevname=1",
+	"console=tty0",
+	"console=ttyS0,115200",
+	"crashkernel=[0-9]*M",
+	"ds=nocloud-net;s=http://",
```
rustydb, line 230: I'd just check for `ds=` because this varies too much.

- PIT boots use `ds=nocloud-net;s=http://${pit-ip}:8888/;h=${hostname}`
- BSS boots use ` ds=nocloud-net;s=http://${bss-ip}:8888/` (for some reason the `h=` bit is not added)

Actually this makes me think we should fix CSI so that it adds the `h=` piece.. I just realized that's a gap.

goss-testing/scripts/data-validator.py
```diff
@@ -192,103 +206,98 @@ def are_hosts_sane(data, desired_keys):
...
+	"rd.net.timeout.ipv6dad=0",
+	"rd.peerdns=0",
+	"rd.retry=10",
+	"rd.shell",
+	"rd.skipfsck",
+	"rd.writable.fsimg=0",
+	"root=live:LABEL=SQFSRAID",
+	"rootfallback=LABEL=BOOTRAID",
+	"transparent_hugepage=never"
+	]
```
rustydb, line 272: The most important ones are these:

- `transparent_hugepage=never`
- `rd.peerdns=0`
- `rd.multipath=0`
- `psi=1`
- `rd.live.*`
- `rd.md.waitclean=1`
- `biosdevname=1`
- `ifname=mgmt*`
- `rd.luks{,=0}`

The rest are important, but the ones listed above will cause the node to boot into a very undesirable state if they're not set properly.

rustydb (approved):

goss-testing/scripts/data-validator.py
```diff
@@ -192,103 +206,98 @@ def are_hosts_sane(data, desired_keys):
 
 
 def validate_ntp(data):
```
rustydb, line 208: It's just me, but since we're using Python3 we can use annotations.

```python
def validate_ntp(data):
```

vs.

```python
# define input types and return type
def validate_ntp(data: dict) -> int
```

So this is a nit, but if you wanted to annotate it does help maintainers see/read what's happening.

Common return types:
```python
def foo(bar: list) -> None
def foo(**kwargs) -> Bool
```

rustydb (changes requested):

goss-testing/scripts/data-validator.py
mharding-hpe: Plus we have other tests to look for this already, I think

rustydb: Pretty sure upgrades validate this, or at least pave over and make it the desired value.

goss-testing/suites/ncn-healthcheck-master.yaml
```diff
@@ -34,6 +34,7 @@ gossfile:
   ../tests/goss-default-gateway-points-to-CMN-gateway.yaml: {}
   ../tests/goss-dns-nodes.yaml: {}
   ../tests/goss-interface-has-ip.yaml: {}
+  ../tests/goss-bss-basecamp-valid.yaml: {}
```
mharding-hpe, line 37: We don't want to add it to this suite. It's already in the corresponding -single.yaml suite, so that it will only run once.
```suggestion
```

goss-testing/suites/ncn-preflight-tests.yaml
```diff
@@ -27,6 +27,7 @@ gossfile:
   ../tests/goss-command-available.yaml: {}
   ../tests/goss-basecamp-json-ncns.yaml: {}
   ../tests/goss-bss-basecamp-valid.yaml: {}
+  ../tests/goss-bss-basecamp-valid.yaml: {}
```
mharding-hpe, line 30: This is a duplicate of the previous line.
```suggestion
```

mharding-hpe (changes requested):

goss-testing/scripts/data-validator.py
jpdavis-prof: It shouldn't be listed twice. I'll remove it entirely.

goss-testing/scripts/data-validator.py
jpdavis-prof: Got it. I'll just look for the prefix.

goss-testing/scripts/data-validator.py
mharding-hpe: I wouldn't worry about implementing this for today, necessarily. The most urgent thing is to fix up the test to avoid the current false failures, before we start seeing them on live systems.

goss-testing/scripts/data-validator.py
jpdavis-prof: I'll just look for **metal.server=**

goss-testing/scripts/data-validator.py
jpdavis-prof: **ds=** it is.

goss-testing/scripts/data-validator.py
jpdavis-prof: So should I only look for the items in your bulleted list? Like, that's the comprehensive list?

goss-testing/scripts/data-validator.py
jpdavis-prof: I'll save this for later.

mharding-hpe (approved):

goss-testing/scripts/data-validator.py
```diff
@@ -192,103 +206,96 @@ def are_hosts_sane(data, desired_keys):
...
 
-  if score_params(worker): err = 1
-  if score_params(storage): err = 1
-  if score_params(management): err = 1
-  return err
+      elif re.search("ncn-s0", hostname):
+        mod_params = ncn_params
+        mod_params = mod_params + storage_params
+
+      elif re.search("ncn-m0", hostname):
```
rustydb, line 288: Do we want the `0` in both of these? It's unlikely we'll have a system with 100 storage or master nodes, but it might still be worth fixing this.

rustydb (changes requested):

goss-testing/scripts/data-validator.py
```diff
@@ -192,104 +206,97 @@ def are_hosts_sane(data, desired_keys):
...
+    if "params" in blob and blob['cloud-init']['user-data'] is not None:
+      hostname = blob['cloud-init']['user-data']['hostname']
 
-  unique_param_strings = set(all_param_strings)
-  if len(unique_param_strings) > 1:
-    for blob in node_class:
-      for params in unique_param_strings:
-        if params == blob['params-stripd']:
-          blob['valid'] = False
+      if re.search("ncn-w00", hostname):
```
rustydb, line 280: Again here, should just look for `ncn-w`, `ncn-m`, and `ncn-s`.

bklei (approved):

rustydb (approved):

PR #376, opened 2022-08-25
CASMINST-5088: Scripts to validate kyverno configuration.

pradeepkumargl:
## Summary and Scope

Scripts to validate kyverno configuration.

### Tested on:

Tested on drax running csm 1.3

### Test description:

Manually tested the scripts with -p parameter and the logs are as below.

Kyverno pods validation:

ncn-m001:/tmp # ./k8s_kyverno_pods_running.sh -p
PASS
ncn-m001:/tmp #

Kyverno Policy report validation:

ncn-m001:/tmp # ./k8s_kyverno_polr_list.sh -p
Error: There are 54 failures in policy report.
Look into the Kyverno policy management document to address the failures/warnings/errors/skipped
FAIL
ncn-m001:/tmp #

## Risks and Mitigations

These are standalone scripts used to troubleshoot Kyverno.

### DIFF
--- a/goss-testing/scripts/k8s_kyverno_pods_running.sh
+++ b/goss-testing/scripts/k8s_kyverno_pods_running.sh
@@ -53,20 +53,23 @@ error_flag=0
 
 # checks if all kyverno pods are running. Total 3 pods.
 
-running_pods=$(kubectl get poda -n kyverno -o json | jq '[.items[] | select(.metadata.labels.app == "kyverno").status.containerStatuses[0].state.running] | length') || err_exit 10 "Command pipeline failed with return code $?: kubectl get pods -n kyverno -o json | jq '[.items[] | select(.metadata.labels.app == "kyverno").status.containerStatuses[0].state.running] | length'"
-if [[ $running_pods != 3 ]] && [[ $print_results -eq 1 ]]
+running_pods=$(kubectl get poda -n kyverno -o json | jq '[.items[] | select(.metadata.labels.app == "kyverno").status.containerStatuses[0].state.running] | length') || err_exit 10 "Command pipeline failed with return code $?: kubectl get pods -n kyverno -o json | jq '[.items[] | select(.metadata.labels.app == \"kyverno\").status.containerStatuses[0].state.running] | length'"
+if [[ $running_pods != 3 ]]
 then
-        error_flag=1;
+    error_flag=1
+    if [[ $print_results -eq 1 ]]
+    then
         echo "Error: $running_pods out of 3 pods are running."
         echo "For high availability the recommended Kyverno pod replica count is 3."
         echo "Check the logs, restart Kyverno and ensure all 3 Kyverno pods are running."
-
+    fi
 fi
-if [[ $running_pods != 3 ]]
+
+if [[ $error_flag -eq 0 ]]
 then
-        error_flag=1;
+    echo "PASS"
+    exit 0
 fi
 
-if [[ $error_flag -eq 0 ]]; then echo "PASS"; exit 0;
-else echo "FAIL"; exit 1;
-fi
+echo "FAIL"
+exit 1
--- a/goss-testing/scripts/k8s_kyverno_pods_running.sh
+++ b/goss-testing/scripts/k8s_kyverno_pods_running.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+#
+# MIT License
+#
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+print_results=0
+while getopts ph stack
+do
+    case "${stack}" in
+          p) print_results=1;;
+          h) echo "usage: k8s_kyverno_pods_running.sh           # Only print 'PASS' upon success"
+             echo "       k8s_kyverno_pods_running.sh -p        # Print all results and errors if found. Use for manual check."
+             exit 3;;
+         \?) echo "usage: k8s_kyverno_pods_running.sh           # Only print 'PASS' upon success"
+             echo "       k8s_kyverno_pods_running.sh -p        # Print all results and errors if found. Use for manual check."
+             exit 3;;
+    esac
+done
+
+error_flag=0
+
+# checks if all kyverno pods are running. Total 3 pods.
+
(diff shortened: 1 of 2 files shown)


mharding-hpe:
Should there also be Goss tests added to use these new scripts? And if so, should those tests be added to some of the test suites?

mharding-hpe (changes requested):
Should there also be Goss tests added to use these new scripts? And if so, should those tests be added to some of the test suites?

Also, I see in the PR template that the second test failed. Have you been able to run the test successfully anywhere? If a test always fails, it either indicates a CSM bug that we need to fix, or it indicates that our test is being too picky. For example, I notice that that test fails if there are skips or warnings, not just errors. Is that appropriate? I would at least suggest adding command line flags so that the user can control whether it fails for those or not.

pradeepkumargl:
> Should there also be Goss tests added to use these new scripts? And if so, should those tests be added to some of the test suites?

These scripts are added only to run manually. We will add goss tests in next release.

pradeepkumargl:
> Should there also be Goss tests added to use these new scripts? And if so, should those tests be added to some of the test suites?
> 

These scripts are added only to run manually. We will add goss tests in next release.

> Also, I see in the PR template that the second test failed. Have you been able to run the test successfully anywhere? If a test always fails, it either indicates a CSM bug that we need to fix, or it indicates that our test is being too picky. For example, I notice that that test fails if there are skips or warnings, not just errors. Is that appropriate? I would at least suggest adding command line flags so that the user can control whether it fails for those or not.

The test case has been tested and it is working. The failure message is that there are failures w.r.t  kyverno which is captured in policy report on the system it was executed. It has been picked by test case and displayed correctly. Yes, it should pick failures/errors/warnings/skipped all the entries in policy report in case of failure.

goss-testing/scripts/k8s_kyverno_pods_running.sh
```diff
@@ -0,0 +1,59 @@
...
+        echo "For high availability the recommended Kyverno pod replica count is 3."
+        echo "Check the logs, restart Kyverno and ensure all 3 Kyverno pods are running."
+
+fi
+if [[ $running_pods != 3 ]]
+then
+        error_flag=1;
+fi
+
+if [[ error_flag -eq 0 ]]; then echo "PASS"; exit 0;
```
mharding-hpe, line 57:
```suggestion
if [[ ${error_flag} -eq 0 ]]; then echo "PASS"; exit 0;
```

goss-testing/scripts/k8s_kyverno_polr_list.sh
```diff
@@ -0,0 +1,86 @@
...
+then
+        if [[ $print_results -eq 1 ]]
+        then
+                echo "Error: Policy report shows $skipped policies are skipped."
+                echo "Look into the Kyverno policy management document to address the failures/warnings/errors/skipped"
+        fi
+        error_flag=1;
+fi
+
+if [[ error_flag -eq 0 ]]; then echo "PASS"; exit 0;
```
mharding-hpe, line 84:
```suggestion
if [[ $error_flag -eq 0 ]]; then echo "PASS"; exit 0;
```

goss-testing/scripts/k8s_kyverno_polr_list.sh
```diff
@@ -0,0 +1,86 @@
...
+             echo "       k8s_kyverno_polr_list.sh -p        # Print all results and errors if found. Use for manual check."
+             exit 3;;
+    esac
+done
+
+error_flag=0
+
+# checks if kyverno policy reports are available
+
+failures=$(kubectl get polr -A | awk '(NR>1)' | awk '{sum+=$4} END {print sum}')
```
mharding-hpe, line 43: This more explicitly sums the failure numbers, rather than relying on the field ordering of kubectl.

```suggestion
failures=$(kubectl get polr -A -o json | jq '[.items[].summary.fail] | add')
```

goss-testing/scripts/k8s_kyverno_polr_list.sh
```diff
@@ -0,0 +1,86 @@
...
+warnings=$(kubectl get polr -A | awk '(NR>1)' | awk '{sum+=$5} END {print sum}')
```
mharding-hpe, line 44:
```suggestion
warnings=$(kubectl get polr -A -o json | jq '[.items[].summary.warn] | add')
```

goss-testing/scripts/k8s_kyverno_polr_list.sh
```diff
@@ -0,0 +1,86 @@
...
+errors=$(kubectl get polr -A | awk '(NR>1)' | awk '{sum+=$6} END {print sum}')
+skipped=$(kubectl get polr -A | awk '(NR>1)' | awk '{sum+=$7} END {print sum}')
```
mharding-hpe, line 46:
```suggestion
skipped=$(kubectl get polr -A -o json | jq '[.items[].summary.skip] | add')
```

goss-testing/scripts/k8s_kyverno_polr_list.sh
```diff
@@ -0,0 +1,86 @@
...
+    esac
+done
+
+error_flag=0
+
+# checks if kyverno policy reports are available
+
+failures=$(kubectl get polr -A | awk '(NR>1)' | awk '{sum+=$4} END {print sum}')
+warnings=$(kubectl get polr -A | awk '(NR>1)' | awk '{sum+=$5} END {print sum}')
+errors=$(kubectl get polr -A | awk '(NR>1)' | awk '{sum+=$6} END {print sum}')
```
mharding-hpe, line 45:
```suggestion
errors=$(kubectl get polr -A -o json | jq '[.items[].summary.fail] | add')
```

goss-testing/scripts/k8s_kyverno_polr_list.sh
```diff
@@ -0,0 +1,86 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+print_results=0
```
mharding-hpe, line 25: You may want to run with "set -o pipefail", so you can test to make sure that the kubectl commands are passing. If the kubectl commands fail, then most likely the counter variables will have no values set, which currently would be interpreted by this script as there being no errors. Therefore I recommend setting pipefail, and then checking the exit codes of the kubectl command pipelines.

goss-testing/scripts/k8s_kyverno_pods_running.sh
```diff
@@ -0,0 +1,59 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+print_results=0
```
mharding-hpe, line 25: I recommend setting -o pipefail, and then testing the return code of the kubectl command being called. If it fails, then it currently would still cause the script to fail, but it would be nicer if the script could accurately report that the failure was because the kubectl command itself failed.

mharding-hpe (approved):

rgross727 (approved):

goss-testing/scripts/k8s_kyverno_pods_running.sh
```diff
@@ -0,0 +1,66 @@
...
+          h) echo "usage: k8s_kyverno_pods_running.sh           # Only print 'PASS' upon success"
+             echo "       k8s_kyverno_pods_running.sh -p        # Print all results and errors if found. Use for manual check."
+             exit 3;;
+         \?) echo "usage: k8s_kyverno_pods_running.sh           # Only print 'PASS' upon success"
+             echo "       k8s_kyverno_pods_running.sh -p        # Print all results and errors if found. Use for manual check."
+             exit 3;;
+    esac
+done
+
+# checks if all kyverno pods are running. Total 3 pods.
```
mharding-hpe, line 42:
```suggestion
# Checks if all kyverno pods are running. There should be a total of 3 running pods.
```

PR #400, opened 2022-09-21
CASMINST-5371: Add logging to k8s_psp_enabled Goss test

mharding-hpe:
## Summary and Scope

Make the following modifications to the k8s_psp_enabled test:
- Change its Goss name field from "k8s_nodes_ready" to "k8s_psp_enabled"
- Move its checking logic into a separate shell script file (k8s_psp_enabled_check.sh)
- Enhance the checking logic by more carefully checking for command failures and unexpected situations
- Provide additional output when performing the checks, to make it clear what exactly happened
- Enable logging of the test using the logrun tool

Despite that being a long-ish list of changes, fundamentally it is still the same test using the same core logic.

If there is approval, I'll backport this into 1.3 as well. Otherwise it will be 1.4 only.

## Issues and Related PRs

This is part of an ongoing series of tickets in which I am adding test logging to those tests which still lack it.

## Testing

I ran the new test (both the script by itself and the updated Goss test) on rocket and verified that they worked as expected.

## Risks and Mitigations

Very low risk. It's still a pretty simple test. The logging is being done using the existing logrun tool.

## Pull Request Checklist

- [X] License file intact
- [X] Target branch correct

### DIFF
--- a/goss-testing/scripts/k8s_psp_enabled_check.sh
+++ b/goss-testing/scripts/k8s_psp_enabled_check.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 #
 # MIT License
 #
--- a/goss-testing/scripts/k8s_psp_enabled_check.sh
+++ b/goss-testing/scripts/k8s_psp_enabled_check.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+#
+# MIT License
+#
+# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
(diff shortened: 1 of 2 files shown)


rustydb (approved):

goss-testing/scripts/k8s_psp_enabled_check.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
rustydb, line 1: nit: would recommend just using `#!/bin/bash`, it is very nuance to have more than 1 BASH shell installed.

rustydb (approved):

goss-testing/scripts/k8s_psp_enabled_check.sh
```diff
@@ -0,0 +1,75 @@
...
+    # Check for string we expect to see if PodSecurityPolicy is enabled. This grep command will fail if the string is absent, causing the test to fail.
+    grep -E '\-\-enable-admission-plugins=.*PodSecurityPolicy' "${TMPFILE}"
+done
+
+# We expect to have checked at least one pod (probably the actual expected number is higher, but for the purposes of this test, we just want to make sure we
+# actually checked something
+[[ $COUNT -gt 0 ]]
+
+# Remove the temporary file. On the off chance that this command fails, we do not want to fail the test. Hence the || true.
+rm "${TMPFILE}" || true
```
mharding-hpe, line 71: May as well remove the temporary file before checking the count
```suggestion
# Remove the temporary file. On the off chance that this command fails, we do not want to fail the test. Hence the || true.
rm "${TMPFILE}" || true

# We expect to have checked at least one pod (probably the actual expected number is higher, but for the purposes of this test, we just want to make sure we
# actually checked something
[[ $COUNT -gt 0 ]]
```

bklei (approved):

rgross727 (approved):

PR #403, opened 2022-09-21
CASMINST-5375: Combine duplicated Goss tests (spire-healthcheck, storage-spire-healthcheck)

mharding-hpe:
## Summary and Scope

These two Goss tests are identical except for the remediation procedure in the case of failure. This PR combines them into a single Goss test file, and uses a conditional statement to print the appropriate remediation procedure. The test itself is unchanged. The now redundant storage-spire-healthcheck test is removed, and the suite which included it now uses the merged test.

Having the duplicated tests is mainly bad because it means having to remember to update both if changes ever need to be made. Combining them makes it much easier from a maintenance point of view.

If there is approval, I'll backport this into 1.3 as well. Otherwise it will be 1.4 only.

## Issues and Related PRs

This is part of ongoing series of PRs I'm using to gradually improve the Goss tests.

## Testing

I tested this on surtur, on both a master node and a storage node. The system was healthy and so in both cases the test passed. In order to get Goss to display the description text, I manually modified the test to force it to fail, and verified that the appropriate description text was shown for the given NCN type.

## Risks and Mitigations

Very low risk -- no change to the test itself.

## Pull Request Checklist

- [X] License file intact
- [X] Target branch correct


goss-testing/suites/ncn-afterpitreboot-healthcheck-storage.yaml
```diff
@@ -23,5 +23,5 @@
 #
 gossfile:
   ../tests/goss-boot-parameter-check-no-wipe.yaml: {}
-  ../tests/goss-storage-spire-healthcheck.yaml: {}
+  ../tests/ncn/goss-spire-healthcheck.yaml: {}
```
mharding-hpe, line 26:
```suggestion
  ../tests/goss-spire-healthcheck.yaml: {}
```

rustydb (approved):

bklei (approved):

rgross727 (approved):

PR #410, opened 2022-10-07
VSHA-477 add conditional check for tests that do not apply to vshasta

jacobsalmela:

## Summary and Scope

- Adds conditionals to skip a small subset of tests when a non-metal environment is detected such as vshasta.
- For metal installs, nothing changes besides the addition of a `vshasta: false` variable
- For `vshasta` installs, this value is overridden to `true` so that the irrelevant tests are skipped

Several tests are not applicable in vshasta, so this adds a simple check by checking for 'HPC' in the `/etc/os-release` file, which is not present on the nodes we deploy to GCP.

Using this logic, the tests still run on metal, but are skipped in a vshasta environment

## Issues and Related PRs

* Partially resolves VSHA-477
* Merge before https://github.com/Cray-HPE/livecd-gcp-infrastructure/pull/163

### Tested on:

  * vshasta2
  * `#redbull`

#### on vshasta2

```
# installed the csm-testing and goss-servers rpm from my csm-testing branch
pit:~ # goss -g /opt/cray/tests/install/livecd/suites/livecd-preflight-tests.yaml --vars <(cat /opt/cray/tests/install/livecd/vars/variables-livecd.yaml <(if [ -f /etc/google_system ]; then echo "vshasta: true" | cat -; else echo "vshasta: false" | cat -; fi)) validate
S..S.....SS..SS....SSSS..................................

Failures/Skipped:

Title: BMC Connectivity
Meta:
    desc: Validates connectivity of the BMCs from the PIT node.
    sev: 0
Command: bmc_connectivity: exit-status: skipped

Title: Bond Members Have Correct MTU
Meta:
    desc: Validates that the bond0 interface members have an MTU of 1500. If the test fails, you may get more information by running it manually: /opt/cray/tests/install/livecd/scripts/check_bond_members_have_correct_mtu.sh 1500
    sev: 0
Command: bond_members_have_correct_mtu: exit-status: skipped

Title: Firmware and BIOS versions and baseline
Meta:
    desc: Validates the correct versions of BIOS and firmware; Validates BIOS settings (when available; dependent on vendor). If this test fails, run "/opt/cray/tests/install/livecd/scripts/check_bios_firmware_versions.sh -b" for more information on the failure. On GigaByte NCNs, a value of "null" may indicate that the BIOS/Firmware needs to be reflashed or that a hard AC power cycle is needed.
    sev: 0
Command: firmware_bios_versions: exit-status: skipped
Command: firmware_bios_versions: stdout: skipped

Title: Service 'conman' Running
Meta:
    desc: Checks whether service 'conman' is running.
    sev: 0
Service: conman: enabled: skipped
Service: conman: running: skipped
...
...
```

#### on `#redbuill`

Installed the same rpm, and ran the same command.  Also ran `csi pit validate --livecd-preflight` and verified it worked there as well.

```
Running LiveCD preflight checks (may take a few minutes to complete)...
WARNING: The PITDATA environment variable is not set.
Writing full output to /opt/cray/tests/install/logs/print_goss_json_results/20221012_163111.371052-113298-sxHZdBms/out

Running tests

Checking test results
Only errors will be printed to the screen

Node: redbull-ncn-m001-pit


GRAND TOTAL: 175 passed, 0 failed
```

## Risks and Mitigations

Low.  This adds a conditional on several tests that are not applicable in a vshasta environment.

## Pull Request Checklist

- [ ] Copyrights updated
- [ ] License file intact
- [ ] Target branch correct
- [ ] CHANGELOG.md updated

### DIFF
--- a/goss-testing/tests/common/goss-bond-members-have-correct-mtu.yaml
+++ b/goss-testing/tests/common/goss-bond-members-have-correct-mtu.yaml
@@ -38,4 +38,10 @@ command:
                 "{{$check_bond_members_have_correct_mtu}}" "{{$bond_member_mtu}}"
         exit-status: 0
         timeout: 20000
+        # skip this test on vshasta
+        {{ if eq true .Vars.vshasta }}
+        skip: true
+        {{ else }}
         skip: false
+        {{ end }}
+
--- a/goss-testing/tests/common/goss-services-enabled.yaml
+++ b/goss-testing/tests/common/goss-services-enabled.yaml
@@ -23,6 +23,8 @@
 #
 # Variable set: services_enabled
 service:
+  # set a var for vshasta
+  {{ $vs := .Vars.vshasta}}
   {{range $service := index .Vars "services_enabled"}}
   {{$service}}:
     title: Service '{{$service}}' Enabled
@@ -31,4 +33,12 @@ service:
       sev: 0
     enabled: true
     running: false
+    # skip testing conman on vshasta
+    {{ if eq $service "conman" }}
+    {{ if eq true $vs }}
+    skip: true
+    {{ else }}
+    skip: false
+    {{ end }}
+    {{ end }}
   {{end}}
--- a/goss-testing/tests/common/goss-services-running.yaml
+++ b/goss-testing/tests/common/goss-services-running.yaml
@@ -23,6 +23,8 @@
 #
 # Variable set: services_running
 service:
+  # set a var for vshasta
+  {{ $vs := .Vars.vshasta}}
   {{range $service := index .Vars "services_running"}}
   {{$service}}:
     title: Service '{{$service}}' Running
@@ -31,4 +33,12 @@ service:
       sev: 0
     running: true
     enabled: true
+   # skip testing conman on vshasta
+    {{ if eq $service "conman" }}
+    {{ if eq true $vs }}
+    skip: true
+    {{ else }}
+    skip: false
+    {{ end }}
+    {{ end }}
   {{end}}
--- a/goss-testing/tests/livecd/goss-bmc-check.yaml
+++ b/goss-testing/tests/livecd/goss-bmc-check.yaml
@@ -35,4 +35,9 @@ command:
                 ipmitool chassis status
         exit-status: 0
         timeout: 20000
+        # skip this test on vshasta 
+        {{ if eq true .Vars.vshasta }}
+        skip: true
+        {{ else }}
         skip: false
+        {{ end }}
--- a/goss-testing/tests/livecd/goss-chrony-config.yaml
+++ b/goss-testing/tests/livecd/goss-chrony-config.yaml
@@ -31,3 +31,9 @@ file:
     owner: root
     group: root
     filetype: file
+    # skip this test on vshasta
+    {{ if eq true .Vars.vshasta }}
+    skip: true
+    {{ else }}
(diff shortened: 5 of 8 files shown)


thiagonache (approved):
looks great!

goss-testing/tests/common/goss-bond-members-have-correct-mtu.yaml
```diff
@@ -38,4 +38,10 @@ command:
                 "{{$check_bond_members_have_correct_mtu}}" "{{$bond_member_mtu}}"
         exit-status: 0
         timeout: 20000
+        # skip this test on vshasta (the HPC string is not found in this file)
+        {{ if readFile "/etc/os-release" | regexMatch ".*HPC.*" }}
```
rustydb, line 42: I'd strongly urge you to check for the existence of `/etc/google_system` instead of parsing the `os-release` file.

The metal OS should always be HPC, but in the distant future that could change due to unavoidable circumstances (it shouldn't change, but if for some reason our HPCM integration causes it to change then this test would fail).

rustydb (changes requested):

goss-testing/tests/common/goss-bond-members-have-correct-mtu.yaml
jacobsalmela: I was going to do that, but this:

> NOTE: Goss will error out during during the parsing phase if the file does not exist, no tests will be executed

https://github.com/aelsabbahy/goss/blob/master/docs/manual.md#templates

Perhaps there is a better way to do logic for tests outside of the goss files themselves, but we have several layers of wrapper scripts so I chose this in order to have it work for now.

rustydb: Weird, no way to leverage this in a conditional?:

```
file:
  /etc/google_system:
    exists: true
```

heemstra: Can't you also do straight bash in `command:` ... or does that cause other problems / confuse goss?  (`! test -f /etc/google_system && "{{$logrun}}" -l ...`)

jacobsalmela: I will investigate and report back

jacobsalmela: I think it confuses it, but I will read up some more and report back.

jacobsalmela: This works, but goss doesn't allow combining `--vars` flags.  

```
redbull-ncn-m001-pit:~ # goss -g test.yml --vars <([ -f /etc/google_system ] && echo "vshasta: true" || echo "vshasta: false") validate
Error: found 0 tests, source: test.yml
vshasta $ goss -g test.yml --vars <([ -f /etc/google_system ] && echo "vshasta: true" || echo "vshasta: false") validate
Total Duration: 0.001s
Count: 1, Failed: 0, Skipped: 0
```

I could attempt to expand the above to also read in the other vars files, but that seems even hackier than what I have in this PR

rustydb (approved):

heemstra (approved):

gbaker-hpe (approved):

gbaker-hpe (approved):

PR #414, opened 2022-11-08
VSHA-503 modify tests for use on vshasta

jacobsalmela:

## Summary and Scope

Add vshasta support to some existing tests by adding a conditional.

Adds a `is_vshasta_node` function.

## Issues and Related PRs

* Partially resolves VSHA-503
* Works with https://github.com/Cray-HPE/livecd-gcp-infrastructure/pull/174

### Tested on:

  * `#redbull`
  * Virtual Shasta

### Test description:

Executed the tests on a vshasta environment and saw the tests were skipped.

The function works on a vshasta pit node:

```
pit:~ # is_vshasta_node ; echo $?
+ [[ -f /etc/google_system ]]
+ return 0
+ echo 0
0
pit:~ #
```

a metal image booted to vshasta:

```
ncn-m002:~ # is_vshasta_node ; echo $?
+ [[ -f /etc/google_system ]]
+ grep --color=auto Google
+ lsblk --noheadings -o vendor
+ return 0
+ echo 0
0
```

and a metal node:

```
redbull-ncn-m001-pit:~ # is_vshasta_node ; echo $?
+ [[ -f /etc/google_system ]]
+ grep Google
+ lsblk --noheadings -o vendor
+ return 1
+ echo 1
1
```

## Risks and Mitigations

Low

## Pull Request Checklist

- [ ] Copyrights updated
- [ ] License file intact
- [ ] Target branch correct
- [ ] CHANGELOG.md updated

### DIFF
--- a/goss-testing/automated/run-ncn-tests.sh
+++ b/goss-testing/automated/run-ncn-tests.sh
@@ -48,7 +48,7 @@ function is_pit_node {
 function is_vshasta_node {
     # This is the best check for an image specifically booted to vshasta
     if [[ -f /etc/google_system ]]; then
-      return $?
+      return 0
     # metal images can still be booted on GCP, so check if there are any disks vendored by Google
     elif lsblk --noheadings -o vendor | grep Google >/dev/null; then
       return $?
--- a/goss-testing/automated/run-ncn-tests.sh
+++ b/goss-testing/automated/run-ncn-tests.sh
@@ -613,6 +613,10 @@ function add_local_vars {
 
     local this_node_name this_node_manufacturer var_string node nodes
 
+    if is_vshasta_node; then
+      var_string+="\nvshasta: true\n"
+    fi
+
     # Add local nodename as variable
     this_node_name=$(hostname -s | grep -Eo '(ncn-[msw][0-9]{3}|.*-pit)$')
     var_string="\n\nthis_node_name: \"${this_node_name}\"\n"
--- a/goss-testing/automated/run-ncn-tests.sh
+++ b/goss-testing/automated/run-ncn-tests.sh
@@ -47,17 +47,13 @@ function is_pit_node {
  
 function is_vshasta_node {
     # This is the best check for an image specifically booted to vshasta
-    if [[ -f /etc/google_system ]]; then
-      return 0
-    # metal images can still be booted on GCP, so check if there are any disks vendored by Google
-    elif lsblk --noheadings -o vendor | grep Google >/dev/null; then
-      return $?
-    else
-      # if the above conditions do not match, it is not running on GCP
-      return 1
-    fi
-}
+    [[ -f /etc/google_system ]] && return 0
 
+    # metal images can still be booted on GCP, so check if there are any disks vendored by Google
+    # if not, we conclude that this is not GCP
+    lsblk --noheadings -o vendor | grep -q Google
+    return $?
+}
 
 # Set some default variables here, both because we use them in this library, and also to save
 # scripts from having to set them if they source this library.
@@ -614,19 +610,25 @@ function add_local_vars {
     local this_node_name this_node_manufacturer var_string node nodes
 
     if is_vshasta_node; then
-      var_string+="\nvshasta: true\n"
+        # vshasta
+        var_string+="\nvshasta: true\n"
+
+        # Since we know this is vshasta, we directly set the node manufacturer variable, rather
+        # than the usual call to ipmitool
+        this_node_manufacturer=vshasta
+    else
+        # Not vshasta -- call ipmitool to determine node manufacturer
+        this_node_manufacturer=$(ipmitool mc info | 
+            grep -E "^Manufacturer Name[[:space:]]{1,}:[[:space:]]*[^[:space:]]" |
+            sed -e 's/^Manufacturer Name[[:space:]]*:[[:space:]]*//' -e 's/[[:space:]]*$//')
     fi
+    # Add hardware manufacturer as variable
+    var_string+="\nthis_node_manufacturer: \"${this_node_manufacturer}\"\n"
 
     # Add local nodename as variable
     this_node_name=$(hostname -s | grep -Eo '(ncn-[msw][0-9]{3}|.*-pit)$')
     var_string="\n\nthis_node_name: \"${this_node_name}\"\n"
     
-    # Add hardware manufacturer as variable
-    this_node_manufacturer=$(ipmitool mc info | 
-        grep -E "^Manufacturer Name[[:space:]]{1,}:[[:space:]]*[^[:space:]]" |
-        sed -e 's/^Manufacturer Name[[:space:]]*:[[:space:]]*//' -e 's/[[:space:]]*$//')
-    var_string+="\nthis_node_manufacturer: \"${this_node_manufacturer}\"\n"
-
     # Get NCN list
     nodes=$(get_ncns)
 
--- a/goss-testing/automated/run-ncn-tests.sh
+++ b/goss-testing/automated/run-ncn-tests.sh
@@ -44,6 +44,20 @@ function is_pit_node {
     [[ -f /etc/pit-release ]]
     return $?
 }
+ 
+function is_vshasta_node {
+    # This is the best check for an image specifically booted to vshasta
+    if [[ -f /etc/google_system ]]; then
+      return $?
+    # metal images can still be booted on GCP, so check if there are any disks vendored by Google
+    elif lsblk --noheadings -o vendor | grep Google >/dev/null; then
+      return $?
+    else
+      # if the above conditions do not match, it is not running on GCP
+      return 1
+    fi
+}
+
 
 # Set some default variables here, both because we use them in this library, and also to save
 # scripts from having to set them if they source this library.
--- a/goss-testing/scripts/python/check_for_unused_drives.py
+++ b/goss-testing/scripts/python/check_for_unused_drives.py
@@ -76,17 +76,20 @@ logger.addHandler(stderr_handler)
 hw_type_map = {
     "GIGA-BYTE TECHNOLOGY CO., LTD": "gigabyte",
     "Hewlett Packard Enterprise": "hpe",
-    "Intel Corporation": "intel" }
+    "Intel Corporation": "intel",
+    "vshasta": "vshasta" }
 
 min_osds_per_storage_node = {
     "gigabyte": 12,
     "hpe": 8,
-    "intel": 1 }
+    "intel": 1,
+    "vshasta": 2 }
 
 max_osds_per_storage_node = {
     "gigabyte": 12,
     "hpe": 8,
-    "intel": 0 }
+    "intel": 0,
+    "vshasta": 0 }
 
 def num_storage_nodes(string):
     num = int(string)
--- a/goss-testing/vars/variables-common.yaml
+++ b/goss-testing/vars/variables-common.yaml
@@ -66,3 +66,5 @@ k8s_ceph_csi_secrets:
   - csi-sma-secret
 
 kubectl: "/usr/bin/kubectl"
+# false by default--overriden when vshasta is deployed
+vshasta: false
--- a/start-goss-servers.sh
+++ b/start-goss-servers.sh
@@ -87,21 +87,13 @@ goss_suites_endpoints_ports | while read suite endpoint port ; do
         continue
     fi
 
-    if is_vshasta_node; then 
-      # Use junit for vshasta
-      echo "starting ${endpoint} in background on port ${port} (vshasta detected)"
-      /usr/bin/goss -g "${suite}" --vars "${tmpvars}" serve \
-          --format junit --max-concurrent 4 \
-          --endpoint "/${endpoint}" \
-          --listen-addr "${ip}:${port}" &
-    else
-      # Start Goss server for this entry.
-      echo "starting ${endpoint} in background on port ${port}"
-      /usr/bin/goss -g "${suite}" --vars "${tmpvars}" serve \
-          --format json --max-concurrent 4 \
-          --endpoint "/${endpoint}" \
-          --listen-addr "${ip}:${port}" &
-    fi
+    # Start Goss server for this entry.
+    echo "starting ${endpoint} in background on port ${port}"
+    /usr/bin/goss -g "${suite}" --vars "${tmpvars}" serve \
+        --format "${results_format}" \
+        --max-concurrent 4 \
+        --endpoint "/${endpoint}" \
+        --listen-addr "${ip}:${port}" &
 done
 echo "Goss servers started in background"
 
--- a/start-goss-servers.sh
+++ b/start-goss-servers.sh
@@ -60,8 +60,13 @@ while true ; do
     sleep 5
 done
 
-# for security reasons we only want to run the servers on the HMN network, which is not connected to open Internet
-ip=$(host "$(hostname).hmn" | grep -Po '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
+if is_vshasta_node; then
+  # on vshasta there is only one network and no hmn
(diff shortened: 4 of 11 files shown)


heemstra (approved):

goss-testing/scripts/python/check_for_unused_drives.py
```diff
@@ -76,17 +76,20 @@
 hw_type_map = {
     "GIGA-BYTE TECHNOLOGY CO., LTD": "gigabyte",
     "Hewlett Packard Enterprise": "hpe",
-    "Intel Corporation": "intel" }
+    "Intel Corporation": "intel",
+    "vshasta": "vshasta" }
```
mharding-hpe, line 80: I don't see any other place which would ever set the hardware type to vshasta. Don't you need to modify that code as well?

Also, if this is done, isn't the vshasta variable then redundant? The hardware type could just be checked instead, it seems like, unless I'm missing something here.

goss-testing/vars/variables-common.yaml
```diff
@@ -66,3 +66,5 @@ k8s_ceph_csi_secrets:
   - csi-sma-secret
 
 kubectl: "/usr/bin/kubectl"
+# false by default--overriden when vshasta is deployed
```
mharding-hpe, line 69: I don't see any code which is doing this overriding in this PR.

start-goss-servers.sh
```diff
@@ -60,8 +60,13 @@ while true ; do
     sleep 5
 done
 
-# for security reasons we only want to run the servers on the HMN network, which is not connected to open Internet
-ip=$(host "$(hostname).hmn" | grep -Po '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
+if lsblk --noheadings -o vendor | grep Google; then
```
mharding-hpe, line 63: If that command is going to be the general way to check if this is vshasta or not, then I'd recommend adding it to the shared Bash function library (`automated/run-ncn-tests.sh`) as a function like `is_vshasta`, similar to the `is_pit_node` function. That would make it easier to re-use, and if we ever need to change it, we won't have to hunt down every place we use the command.

mharding-hpe (changes requested):

goss-testing/scripts/python/check_for_unused_drives.py
jacobsalmela: There actually is no ipmitool on vshasta.  We just run a mock version of it, so we just fake it.

https://github.com/Cray-HPE/livecd-gcp-infrastructure/pull/174/files#diff-34e0270d5867cf8e17dc52ef091290e1cf3b721c71c3d9181004129abf8aa7b5R207

goss-testing/vars/variables-common.yaml
jacobsalmela: This var went in in a few spots.  It is false by default in the vars files.  There is a hacky override when we run in vshasta to set it to true.  It is false by default so it does nothing on metal installs

mharding-hpe:
In addition to creating the `is_vshasta` function in [the common function library](https://github.com/Cray-HPE/csm-testing/blob/main/goss-testing/automated/run-ncn-tests.sh), it seems like you'd also want to add code to the `add_local_vars` function in that library to handle any new variables you want to have set.

start-goss-servers.sh
jacobsalmela: We actually have a better `is_vshasta` check, which is to see if `/etc/google_system` exists, but we boot the metal images, which doesn't have that file, so the method of determining if a node is vshasta or not needs some more consideration I think, rather than what I have here.

goss-testing/vars/variables-common.yaml
jacobsalmela: See https://github.com/Cray-HPE/csm-testing/pull/410/files#diff-fd12da3b417be82600b83d8504e8a7a264a13585ce145769ac363277945447ffR102

jacobsalmela: All of the overrides happen in the vshasta codebase.

start-goss-servers.sh
mharding-hpe: Either way, with whatever method you arrive at, it seems like we'd want it to be in a common function for easier reuse, right?

goss-testing/vars/variables-common.yaml
mharding-hpe: Why would you not put it in here, if we already have a method for determining that this is a Vshasta system? Seems easier to maintain that way, rather than requiring PRs in two different repos if wanting to make a change to that part of it. No?

start-goss-servers.sh
jacobsalmela: Yeah, it is not a hard change to make.  I am fixing it up.

goss-testing/vars/variables-common.yaml
jacobsalmela: We certainly could put it in here.  This repo is one of the least developer-friendly repos to contribute to, so that is the main reason I never even considered it.   Also, I already needed to contribute to here, to the csm repo and to csm-rpms in order for my change to show up a few days later, so keeping the vshasta stuff in the vshasta repo meant I could develop it faster. 

I can certainly add it here if that is a gate for you.

mharding-hpe: I would strongly prefer that we have it here. As you've already seen, this repo contains some tests which are not part of any test suite and may or may not still be in use. The same is true of parts of tests and scripts in here -- there are sections which may or may not still be needed. Down the road, if someone notices that we have a lot of vshasta checks, but that there is no place where we ever set it to true, they may conclude that this is another case of something that is no longer in use.

Also, if a change is made to csm-testing that results in the variable files having a different format, or location, or whatever, right now someone making that change would have no way to know that a change would also be required in the vshasta repo. That is another one of the benefits of doing this in the add_local_vars function. It will automatically be kept up to date.

Put another way, I think it's not ideal to have one repo (your vshasta repo) hard-code in assumptions about how another repo (csm-testing) has its stuff internally organized. The automated scripts we provide are really the only interface that should be relied on externally. For example, we don't have anything outside of the repo which directly calls the Goss endpoints, for exactly this reason. There's no promise made that the URLs, port numbers, etc, won't change.

jacobsalmela: Makes sense.  I will make the adjustments

mharding-hpe: For what it's worth, I completely agree with you about this repository not being developer-friendly. That is something we need to improve on.

goss-testing/scripts/python/check_for_unused_drives.py
jacobsalmela: I think we may still want this code.  I think this test should still run on vshasta, rather then being skipped on the `is_vshasta` conditional.

goss-testing/automated/run-ncn-tests.sh
```diff
@@ -44,6 +44,20 @@ function is_pit_node {
     [[ -f /etc/pit-release ]]
     return $?
 }
+ 
+function is_vshasta_node {
+    # This is the best check for an image specifically booted to vshasta
+    if [[ -f /etc/google_system ]]; then
+      return $?
```
mharding-hpe, line 51: Should be the same, but seems more straightforward to not use the variable when we always know it is 0?

```suggestion
      return 0
```

goss-testing/automated/run-ncn-tests.sh
```diff
@@ -44,6 +44,20 @@ function is_pit_node {
...
+    # metal images can still be booted on GCP, so check if there are any disks vendored by Google
+    elif lsblk --noheadings -o vendor | grep Google >/dev/null; then
+      return $?
```
mharding-hpe, line 54: Should be the same, but seems more straightforward to not use the variable when we always know it is 0?

```suggestion
      return 0
```

goss-testing/scripts/python/check_for_unused_drives.py
mharding-hpe: Yeah, I agree. I was saying that if you have the hardware manufacturer variable set to `vshasta`, then you don't also need the other `vshasta` boolean variable. You could just check the manufacturer variable. But it's not a big deal.

start-goss-servers.sh
```diff
@@ -74,12 +79,21 @@ goss_suites_endpoints_ports | while read suite endpoint port ; do
...
     fi
 
-    # Start Goss server for this entry.
-    echo "starting ${endpoint} in background on port ${port}"
-    /usr/bin/goss -g "${suite}" --vars "${tmpvars}" serve \
-        --format json --max-concurrent 4 \
-        --endpoint "/${endpoint}" \
-        --listen-addr "${ip}:${port}" &
+    if is_vshasta_node; then 
+      # Use junit for vshasta
```
mharding-hpe, line 83: This will mean that the other scripts that call the endpoints won't work with vshasta, since they expect it to be in json format. Not sure if that is of concern.

start-goss-servers.sh
```diff
@@ -60,8 +60,13 @@ while true ; do
...
 
-# for security reasons we only want to run the servers on the HMN network, which is not connected to open Internet
-ip=$(host "$(hostname).hmn" | grep -Po '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
+if is_vshasta_node; then
+  # on vshasta there is only one network and no hmn
+  ip=$(host "$(hostname)" | grep -Po '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
+else
+  # for security reasons we only want to run the servers on the HMN network, which is not connected to open Internet
+  ip=$(host "$(hostname).hmn" | grep -Po '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
+fi
```
mharding-hpe, line 69:
```suggestion
if is_vshasta_node; then
    echo "vshasta detected."

    # Use junit for vshasta
    results_format=junit

    # on vshasta there is only one network and no hmn
    ip=$(host "$(hostname)" | grep -Po '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
else
    # for security reasons we only want to run the servers on the HMN network, which is not connected to open Internet
    ip=$(host "$(hostname).hmn" | grep -Po '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
    
    # Return results in JSON format outside of vshasta
    results_format=json
fi
```

start-goss-servers.sh
```diff
@@ -74,12 +79,21 @@ goss_suites_endpoints_ports | while read suite endpoint port ; do
...
+          --endpoint "/${endpoint}" \
+          --listen-addr "${ip}:${port}" &
+    else
+      # Start Goss server for this entry.
+      echo "starting ${endpoint} in background on port ${port}"
+      /usr/bin/goss -g "${suite}" --vars "${tmpvars}" serve \
+          --format json --max-concurrent 4 \
+          --endpoint "/${endpoint}" \
+          --listen-addr "${ip}:${port}" &
+    fi
```
mharding-hpe, line 96:
```suggestion
    # Start Goss server for this entry.
    echo "starting ${endpoint} in background on port ${port}"
    /usr/bin/goss -g "${suite}" --vars "${tmpvars}" serve \
        --format "${results_format}" \
        --max-concurrent 4 \
        --endpoint "/${endpoint}" \
        --listen-addr "${ip}:${port}" &
```

mharding-hpe (changes requested):

start-goss-servers.sh
jacobsalmela: I do not think so... as it is now, our deployment is the only thing that runs the tests.

goss-testing/scripts/python/check_for_unused_drives.py
jacobsalmela: True.  It is redundant, but it is a very small component of vshasta and mostly superfluous because we just use it to fake out some other scripts that depend on it, but we actually have the data elsewhere

goss-testing/vars/variables-common.yaml
jacobsalmela: testing after the recent changes

jacobsalmela: It worked

start-goss-servers.sh
jacobsalmela: Fixed.  More relevant info here: https://github.com/Cray-HPE/csm-testing/pull/414#discussion_r1018211533

goss-testing/automated/run-ncn-tests.sh
mharding-hpe: This was marked resolved but it looks to me like it is not? Or is github messing with me again? Sometimes it likes to show me stale PR data, I think.

goss-testing/vars/variables-common.yaml
mharding-hpe: The code to set this in the `add_local_vars` function should still be added, using the `is_vshasta_node` function that you created.

mharding-hpe (changes requested):
Some of the comments were marked resolved but do not appear to be. I've marked them unresolved and added comments.

goss-testing/automated/run-ncn-tests.sh
jacobsalmela: Oh yeah, I co-authored the commit with you.  https://github.com/Cray-HPE/csm-testing/pull/414/commits/be2345b6a3bea4ce0080b75151b5e46be7d259e0.

The same thing happens to me, but I took all your suggestions if you check the commits tab.

goss-testing/vars/variables-common.yaml
jacobsalmela: oops. I'll look

jacobsalmela: ok check now

goss-testing/automated/run-ncn-tests.sh
mharding-hpe: I don't see the above suggestion in the commit tag.

mharding-hpe (changes requested):
I am still seeing the return $? on line 54 of `goss-testing/automated/run-ncn-tests.sh`, even if I view it directly from a git clone of the branch.

I also still see a change needed in the `add_local_vars` function in that same file.

To save time, I am going to make a PR against your PR branch with the changes I still think should be made. You can merge it into your PR branch if you agree.

goss-testing/automated/run-ncn-tests.sh
jacobsalmela: https://github.com/Cray-HPE/csm-testing/pull/414/commits/afc2c42471a264a846a023b26ed51e9fddb9954e

mharding-hpe (changes requested):
Please take a look at https://github.com/Cray-HPE/csm-testing/pull/416

That addresses my remaining concerns,

If you don't have authority to merge it into this branch, just review it and approve it, and I can merge it in here.

goss-testing/vars/variables-common.yaml
mharding-hpe: I see the code for the vshasta variable in there, but it looks like it still sets the node manufacturer using ipmitool, which I believe you said will not work for vshasta.

I made this PR against your PR branch:
https://github.com/Cray-HPE/csm-testing/pull/416

It will address that issue, and also make some small formatting changes to make the new code more closely match the style of the existing functions.

jacobsalmela:
Pulled it in

mharding-hpe (approved):

mharding-hpe:
> Pulled it in
Perfect. I've approved, but looks like you may need to squash the commits now because of the merge commit that snuck in there.

jacobsalmela:
Removed the merge commit.  I think we're good now

mharding-hpe:
> Removed the merge commit. I think we're good now

Cool. I think you need someone on PET team to approve this before it can merge. Or anyway, I don't have authority to do it without that. Maybe go hit up #casm-pet channel and beg. :)

jacobsalmela:
Looks like I cannot merge it yet either

bklei (approved):

PR #427, opened 2023-01-13
CASMINST-5793: print_goss_json_results: Do not use scientific notation when logging test duration in seconds; minor bug fixes

mharding-hpe:
## Summary and Scope

Sumith R B requested that print_goss_json_results not use scientific notation for test durations in the log file consumed by grok exporter. This PR makes that change, and also fixes some small bugs that I found while coding and testing this change.

## Issues and Related PRs

[CASMINST-5793](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-5793)

## Testing

The change is not large, but the code path is one which is hit a lot, since every test result that is parsed is also logged. However, because of this, it was easy to do extensive testing of this. I did so on wasp. To the end user, there is no change. The only external difference is in the format of the grok exporter log file.

Sumith also tested the new code on mug and verified that it produced logs that could be consumed by grok exporter.

## Risks and Mitigations

If this is omitted, then not only will the grok exporter not be able to consume the logs properly, but the bugs I fixed will remain. They were in error paths in the code, but they would make problems harder to debug when they occur (since they relate to the script logging exceptions). I think the risk of not including this is worse than the risk of including it.

## Pull Request Checklist

- [X] Copyrights updated
- [X] License file intact
- [X] Target branch correct

### DIFF
--- a/goss-testing/automated/python/lib/grok_exporter_logger.py
+++ b/goss-testing/automated/python/lib/grok_exporter_logger.py
@@ -2,7 +2,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022-2023 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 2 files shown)


mharding-hpe:
The shellcheck errors are for files not being modified by this PR

sumithrb (approved):

ndavidson-hpe (approved):

PR #454, opened 2023-03-15
Fix shellcheck errors

mtupitsyn:
## Summary and Scope

Fix shellcheck errors.

### DIFF
--- a/goss-testing/automated/run-ncn-tests.sh
+++ b/goss-testing/automated/run-ncn-tests.sh
@@ -659,6 +659,7 @@ function add_local_vars {
     var_string+="\n${is_vshasta}\n"
 
     echo -e "${var_string}" >> "$1"
+    # shellcheck disable=SC2320
     return $?
 }
 
--- a/goss-testing/scripts/check_interface_mac_matches_bond.sh
+++ b/goss-testing/scripts/check_interface_mac_matches_bond.sh
@@ -77,6 +77,7 @@ if ! eval ip addr show dev "$INTERFACE" 1>/dev/null; then
     break
   # otherwise, it's a legit failure
   else
+    # shellcheck disable=SC2319
     error 30 $? "ip addr show dev $INTERFACE"
   fi
 # if the interface does exist, continue checking if it matches bond0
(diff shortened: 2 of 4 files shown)


bklei (approved):

goss-testing/automated/run-ncn-tests.sh
```diff
@@ -659,6 +659,7 @@ function add_local_vars {
     var_string+="\n${is_vshasta}\n"
 
     echo -e "${var_string}" >> "$1"
+    # shellcheck disable=SC2320
```
kburns-hpe, line 662: This is a valid issue. Why are we ignoring it?

goss-testing/scripts/check_interface_mac_matches_bond.sh
```diff
@@ -77,6 +77,7 @@ if ! eval ip addr show dev "$INTERFACE" 1>/dev/null; then
     break
   # otherwise, it's a legit failure
   else
+    # shellcheck disable=SC2319
```
kburns-hpe, line 80: This is a valid issue. Why are we ignoring it?

kburns-hpe (changes requested):

goss-testing/scripts/check_interface_mac_matches_bond.sh
mtupitsyn: Well, I just don't know how to get `$?` from `if ! eval ip addr show dev "$INTERFACE" 1>/dev/null; then`. Even if you put `rc=$?` right after it; it will be return code from if statement, not from `eval`. Any ideas?

goss-testing/automated/run-ncn-tests.sh
mtupitsyn: My understanding of this code is that we indeed using return code of `echo` command as return code of this function. For example, if `"$1"` is a read-only file, it will be non-zero.

kburns-hpe: Ok, that makes sense. thanks.

goss-testing/scripts/check_interface_mac_matches_bond.sh
mtupitsyn: Yep, this is what was implemented in https://github.com/Cray-HPE/csm-testing/pull/456

PR #473, opened 2023-03-24
CASMINST-6130: Fix HOSTNAME var in start-goss-servers.sh

johren-hpe:
## Summary and Scope

The HOSTNAME var being set in start-goss-servers.sh doesn't handle the hostname of the pit in vshasta because it is just "pit".   I removed the dash before the pit because it is not needed.   It was only there because I borrowed the regex from run-ncn-tests.sh. 

## Issues and Related PRs

* Resolves [CASMINST-6130](https://jira-pro.its.hpecorp.net:8443/browse/CASMINST-6130)

### Tested on:

  * `fanta`
  * Virtual Shasta - `cilium`

### Test description:

On both fanta's m002 and cilium's pit node, I started the goss-servers service and made sure that there were no errors.  I also ran `curl http://ncn-m002.hmn:8997/ncn-healthcheck-master` on fanta to make sure the endpoint still works.

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [ ] CHANGELOG.md updated

### DIFF
--- a/start-goss-servers.sh
+++ b/start-goss-servers.sh
@@ -35,7 +35,7 @@ fi
 export KUBECONFIG=/etc/kubernetes/admin.conf
 
 # necessary for test that need to know the current hostname
-HOSTNAME=$(hostname -s | grep -Eo '(ncn-[msw][0-9]{3}|.*-pit)$')
+HOSTNAME=$(hostname -s | grep -Eo '(ncn-[msw][0-9]{3}|.*pit)$')
 export HOSTNAME
 
 # During the NCN image build, this service is started, even though the csm-testing RPM is not installed. In that
--- a/start-goss-servers.sh
+++ b/start-goss-servers.sh
@@ -35,7 +35,7 @@ fi
 export KUBECONFIG=/etc/kubernetes/admin.conf
 
 # necessary for test that need to know the current hostname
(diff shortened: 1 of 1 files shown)


start-goss-servers.sh
```diff
@@ -35,7 +35,7 @@ fi
 export KUBECONFIG=/etc/kubernetes/admin.conf
 
 # necessary for test that need to know the current hostname
-HOSTNAME=$(hostname -s | grep -Eo '(ncn-[msw][0-9]{3}|.*-pit)$')
+HOSTNAME=$(hostname -s | grep -Eo '(ncn-[msw][0-9]{3}|.*pit)$')
```
mtupitsyn, line 38: Maybe just replace `grep` with `sed` - so that if there's unexpected hostname, it remains unchanged?

mtupitsyn (approved):

jacobsalmela (approved):

start-goss-servers.sh
johren-hpe: Okay, that will need retesting.   I'll circle back with this suggestion next week.

mtupitsyn: Makes sense. Thank you for returning to this topic!

mtupitsyn (approved):

PR #482, opened 2023-04-06
Fix Jenkins

rustydb:
This change does a few things.

1. It fixes a race condition in the `matrix` where one axis overrides the other axis' delivery point. We've observed the SP3 build publishing RPMs into the SP4 build on more than one occasion (see https://jenkins.algol60.net/blue/organizations/jenkins/Cray-HPE%2Fcsm-testing/detail/v1.15.43/2/pipeline/105#step-205-log-2). This is often "fixed" by a re-run, but that has no guarantees of working.
1. This lightens the load on Jenkins by restricting the build to a single GCP agent and spawning a container atop that agent for each axis. Right now multiple hosts are requested, this adds needless demand to Jenkins and is slower.
1. This updates the `Makefile` and `csm-testing.spec` to accommodate the above changes
1. Adds `dependabot.yaml` for auto-updating GitHub workflows

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    # Look for `.github/workflows` in the `root` directory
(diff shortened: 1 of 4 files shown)


rustydb:
The race condition mentioned exists in all the release branches that have a `matrix` defined, and it was thought to have been fixed in main. This PR just normalizes the fix to a Jenkinsfile schema that I know 100% fixes the race condition.

heemstra (approved):

rustydb:
https://jira-pro.its.hpecorp.net:8443/browse/MTL-2088

PR #500, opened 2023-06-02
MTL-2167 Handle new ipxe files from MTL-2139

rustydb:
`ipxe.efi` is now two files, `ipxe.x86_64.efi` and `ipxe.arm64.efi`.


heemstra (approved):

denniswalker:
rusty swears he ran this code

PR #512, opened 2023-07-11
CASMINST-6523: Modify upgrade stage 0 iPXE check; create scripts to run Goss tests during CSM upgrades

mharding-hpe:
## Summary and Scope

This PR makes the following changes/additions:

* Creates a new iPXE check script that will pass if the CSM 1.4 style iPXE pod is running or if the CSM 1.5 style x86-64 iPXE pod is running (it will not pass if only the aarch64 pod is running, because that won't work for NCN boots).
* Modifies the iPXE test run during stage 0 of CSM upgrades to use this new script. This is the only situation where we run the iPXE test that it may be the old or new style of deployment. All other cases will expect only the new style of deployment.
* Because of this change, the `ncn-patch-upgrade-preflight-tests.yaml` suite is no longer needed. It isn't being used (it was intended to be used to handle the iPXE situation for CSM 1.5.1 upgrades) so this PR deletes it.
* I noticed that the upgrade scripts are calling Goss directly to run the test suites. This isn't a big deal except I noticed they also have code to manually fix the variables file. This is something that is already handled in the shared function library in the `csm-testing` repo, and can easily be leveraged to avoid having to make those kind of changes in the `docs-csm` repo. This is better because the way things are now, the upgrade scripts could break if internal `csm-testing` repo changes are made, and the person making those changes would likely not realize that this is happening in `docs-csm`.

## Issues and Related PRs

* Resolves [CASMINST-6523](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6523)/[CASMINST-6541](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6541)
* For CSM 1.6 we can go back to having a monolithic iPXE check, since the source version will always be using the new style of deployment. I have [CASMINST-6522](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6522) open for that, and will implement it once we branch `csm-testing`.

## Testing

I tested the new iPXE test (both shell script and Goss test) on wasp (CSM 1.4) and ashton (CSM 1.5) to verify that it works as expected.
I also tested the new automated shell scripts on wasp to make sure that they behaved as expected.

## Risks and Mitigations

Low risk.

## Pull Request Checklist

- [X] Copyrights updated
- [X] License file intact
- [X] Target branch correct

### DIFF
--- a/goss-testing/automated/ncn-post-csm-service-upgrade-tests
+++ b/goss-testing/automated/ncn-post-csm-service-upgrade-tests
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+#
+# MIT License
+#
+# (C) Copyright 2023 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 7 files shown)


mharding-hpe:
Note that git thinks I have renamed a file, but in truth I deleted one file and created another one. It thinks it was a rename because of the algorithm git uses to make such determinations (namely that the new file is > 50% similar to the old file, I believe). Because both files are dominated by their copyright headers, it's not surprising, but do not be deceived.

cdelatte-hpe (approved):

PR #559, opened 2023-11-30
CASMINST-6730: Add Goss test to verify that all Kubernetes NCNs are at the same level

mharding-hpe:
## Summary and Scope

In the course of investigating [CASMTRIAGE-6371](https://jira-pro.it.hpe.com:8443/browse/CASMTRIAGE-6371) it was discovered that one of the worker NCNs was overlooked during the CSM upgrade. However, this was only caught essentially by accident. No Goss test exists to catch this kind of issue.

This PR adds a test which consults Kubernetes and compares several values for all of the master and worker NCNs. If they do not match, the test fails. For most of these fields, it expects them to match across both masters and workers. For one field (kernel version), it just expects masters to match masters and workers to match workers. This is to account for the fact that worker images are customized differently than master images.

This test is added to the suite which runs only on a single master node, since it is comparing data for all Kubernetes NCNs, regardless of where it runs. So there is no benefit to running it multiple times.

## Issues and Related PRs

* Inspired by [CASMTRIAGE-6371](https://jira-pro.it.hpe.com:8443/browse/CASMTRIAGE-6371)
* Resolves [CASMINST-6730](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6730)
* Partially resolves [CASMINST-6729](https://jira-pro.it.hpe.com:8443/browse/CASMINST-6729)
* [CSM 1.5 backport PR](https://github.com/Cray-HPE/csm-testing/pull/560)

## Testing

I tested this on mug, starlord, and wasp. For mug and starlord, it correctly reported differences between the NCNs. For wasp, there were no such differences, and the test passed.

## Risks and Mitigations

Low risk.

## Pull Request Checklist

- [X] Target branch correct

### DIFF
--- a/goss-testing/scripts/python/compare_k8s_ncns.py
+++ b/goss-testing/scripts/python/compare_k8s_ncns.py
@@ -1,0 +1,169 @@
+#!/usr/bin/env python3
+#
+# MIT License
+#
+# (C) Copyright 2023 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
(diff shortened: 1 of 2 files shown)


don-bahls-hpe (approved):

kimjensen-hpe (approved):

goss-testing/suites/ncn-kubernetes-tests-master-single.yaml
```diff
@@ -34,6 +34,7 @@
 # corresponding suite file without the -single suffix.
 gossfile:
   ../tests/goss-ceph-csi-k8s-requirements.yaml: {}
+  ../tests/goss-compare-k8s-ncns.yaml: {}
```
leliasen-hpe, line 37: Why did you choose to add it to this suite? Is this suite run after all nodes have been upgraded and never between node upgrades? I see there is a suite that is run [after csm-upgrades](https://github.com/Cray-HPE/csm-testing/blob/2d629b5f632d22b99d0f2d018aabe95c5383571f/goss-testing/automated/run-ncn-tests.sh#L551)

leliasen-hpe (changes requested):

goss-testing/suites/ncn-kubernetes-tests-master-single.yaml
mharding-hpe: I could also add it to the suite for after csm upgrades, but I also wanted it to be part of the regular CSM health checks. That way it would catch other cases where the nodes ended up mismatched (for example, if a node was rebuilt using the wrong image).

I don't think the suite I added it to is run during CSM upgrades, but I will double check now.

mharding-hpe: I searched the `cray-nls` and `docs-csm` repos for all places where this suite is called, and there are no cases where it is called during an upgrade. It is only run during normal CSM health checks (which are supposed to be done before and after upgrades).

leliasen-hpe: Okay, makes sense. Thanks for looking into it!

leliasen-hpe (approved):

PR #596, opened 2024-08-27
Added example test for testing enhancement effort

shunr-hpe:
## Summary and Scope

Created the ct-testing directory. Added the csm-ct.py script for running the goss and ansible tests for the ct testing effort.

CASMINST-6936

### Tested on:

  * Virtual Shasta

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/ct-testing/tests/example-test-ansible-setup.yaml
+++ b/ct-testing/tests/example-test-ansible-setup.yaml
@@ -1,0 +1,8 @@
+---
+- name: node-consistency-setup
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  tasks:
+    - name: setup
+      shell: echo "Running node consistency setup."
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+*.swp
+.idea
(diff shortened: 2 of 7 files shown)


jacobsalmela (commented):
I mentioned this, but I think there is overlap with the **hpc-install-test-pipeline-automation** repo as it uses a wrapper and a config file to run the tests.  

The tests in this repo are already wrapped in a couple layers of wrappers; yours is more flexible, but we also have wrappers in other places, too.  

We should make sure adding a new test is trivial, and same for running it.  We have different scenarios in which it will run:

- nightly vshasta2 builds
- ad-hoc/manual triggers on virtual or physical machines
- ct pipeline that runs on real hardware

The vshasta2 flow will offer the biggest return on investment and might require changes there to get these tests to run in that flow.

I like the utility, and it may have value going forward.  I need to ponder this a bit more.

ct-testing/tests/example-test-ansible-setup.yaml
```diff
@@ -0,0 +1,8 @@
+---
```
jacobsalmela, line 1: I realize these are just examples, but we should also be able to get these installed with this rpm.  I like the idea of having our own spot to work, but we'd eventually need a change to the spec file or move these in with a path that gets installed already.

PR #604, opened 2024-10-17
CASMAUTO-53 add a new script to run tests and make dst results.json

jacobsalmela:
CASMAUTO-53 add a new script that runs all available goss tests on any node type and creates a DST-compatible json

It works by looking for the commands that are running in goss-servers
and reformats them so they can be executed directly by the script, or
potentially a human trying to debug them.  This was needed to get the
JSON output without needing to go through several layers of wrapper
scripts that currently exist.

While it may be possible to have the DST pipeline properly execute the
goss tests with the existing framework and wrapper scripts, it proved to
be challenging and easy to get lost on what calls what.  There is a
strong need for three important things:

1. it should be trivial for a contributor of any skill level to add a
   new test
1. any new tests should be automatically run the next time the rpm is
   installed as part of a build
1. the test output needs to be machine-readable, which is currently
   limited with the existing wrapper scripts and http endpoints (or it
   is non-trivial to modify it to do so)

Here is an example of this complexity I am trying to overcome with this
script:

1. a systemd service is started with `start-goss-servers.sh`, which reads `dat/goss-servers.cfg` and runs `goss serve` for each suite file
1. each suite file has one or more individual test files it will run,
   many of which are in multiple suites

those two steps are logical and a reasonable setup.  when you start
adding in these wrapper scripts, it gets confusing:

1. a comprehensive helper script, `run-ncn-tests.sh` has many functions, one of which sets up temporary variable files, which is used by the `start-goss-servers.sh`
1. with the http endpoints up and running, additional wrapper scripts
   are used to facilitate __calling__ the endpoints and vary in their
   methods to do so
1. in most cases, `run-ncn-tests.sh` is `source`d by the
   `automated/<wrapper>` script, and there is some bash functions, which
   call a python script that will determine the http endpoint urls and
   get appropriate node types
1. in addition to the endpoints from the python wrapper script, the bash
   function also accepts individual test files, so each wrapper script
   can add their own extra tests if needed.  this functionality is
   helpful, but makes it difficult to track down what is being run
1. these endpoints may vary depending on different factors: if it is a
   vshasta node, if it is the pit or not, if kube credentials are
   available or not, so logic exists in all these wrapper scripts to
   calculate that
1. some of that logic can also add a suite instead of a test, so there
   is the potential for individual tests to be run multiple times
1. some of these wrapper scripts run different functions to the same end
   result.  some call `run_goss_tests_print_results`, some call
   `print_goss_json_results`, some do one or the other based on previous
   logic, and some simply call `run_goss_tests` with little other logic,
   one even skips those helper functions and just calls goss directly
   (`ncn-postgres-tests`)
1. in
   https://cray-hpe.github.io/docs-csm/en-15/operations/validate_csm_health/,
   the document mainly calls out
   `/opt/cray/tests/install/ncn/automated/ncn-k8s-combined-healthcheck`,
   which is one of the wrapper scripts, so it isn't immediately clear
   where the others come into play
1. in addition to this automated wrapper script, the document also calls
   out:
    1. some PET wrappers: `/opt/cray/platform-utils/ncnHealthChecks.sh`
    1. some HMS wrappers:
       `/opt/cray/csm/scripts/hms_verification/hsm_discovery_status_test.sh`
       and `/opt/cray/csm/scripts/hms_verification/verify_hsm_discovery.py`
    1. some CMS wrappers: `/usr/local/bin/cmsdev`
    1. a barebones test:
       `/opt/cray/tests/integration/csm/barebonesImageTest`

1. HMS and CSM both have goss test files in this repo, but also use
   their own wrappers, which may do other things I have not investigated
1. in addition to the wrapper scripts, the vars files contain some
   things, but not as many as would be most useful.  for example, the
   `pod_services` has only four services which are checked, but there
   are much more.
1. some of these vars could be dynamically gathered at run time. this is
   out of scope right now for this ticket, but worth mentioning in the
   complexity of things since we have a lot of complexity, but are then
   prone to not testing everything at that expense (such as testing four
   services instead of all of them)
1. some tests only need to run from one node (for example, if they are
   only querying the apis or running a kubectl command that would be the
   same on every node), while others need to run on each node and run
   local commands, so there is likely some overlap and unnecessary
   running of some tests in some places, but again, it is hard to flush
   out where and what exactly is happening

This new script I made adds to the complexity of having yet another
wrapper, but the end goal is to allow the three important bullets I
mentioned previously and to allow the tests to run in the pipeline and
report properly to the DST dashboard.  Since these tests are a gate to a
release, we need to ensure they are actually run properly and reported
properly; right now, they are not.

So this new script jumps in front of the complexity and queries the goss
endpoints, borrows their created variable files and creates a goss
command that can be run by a human or a machine.

This script can run on any node type and will only run the tests it
finds (those that are setup with a goss endpoint).  this allows the
majority of this repo to remain the same and will allow any new changes
people are used to making to automatically be executed when this script
runs.

this script's current puprose is to create the results.json file DST can
use, but it could be grown to replace some wrappers or improve upon
them.

We also need the ability to run and individually trigger smoke, funcitonal, and destructive level tests, both ad-hoc and automated.

So while this script is not perfect right now, it can run on all nodes
and run the existing tests.  this is step one to get the gate in a
better position and allow newly-added tests to run without a change to
this script.

I ran this new script on three node types:

**Master**

```
ncn-m002:~ # ./dst-ct-results.sh
Found 14 suites to run
Running ncn-preflight-tests.yaml...DONE
Running ncn-smoke-tests.yaml...DONE
Running ncn-spire-healthchecks.yaml...DONE
Running ncn-healthcheck-master.yaml...DONE
Running ncn-healthcheck-master-single.yaml...DONE
Running ncn-afterpitreboot-healthcheck-master.yaml...DONE
Running ncn-kubernetes-tests-master.yaml...DONE
Running ncn-kubernetes-tests-master-single.yaml...DONE
Running ncn-afterpitreboot-kubernetes-tests-master-single.yaml...DONE
Running ncn-kubernetes-tests-cluster.yaml...DONE
Running ncn-cms-tests.yaml...DONE
Running ncn-hms-ct-tests.yaml...DONE
Running ncn-post-csm-service-upgrade-tests.yaml...DONE
Running ncn-healthcheck-master-single-post-service-upgrade.yaml...DONE
Aggregated goss results have been saved to: /tmp/tmp.ppyE5WIK3y
DST-and-ct-results-compatible file been saved to: /tmp/api-results.json
ncn-m002:~ # jq '.tests | length' < /tmp/api-results.json
297
```

**Worker**

```
ncn-w002:~ # ./dst-ct-results.sh
Found 10 suites to run
Running ncn-preflight-tests.yaml...DONE
Running ncn-smoke-tests.yaml...DONE
Running ncn-spire-healthchecks.yaml...DONE
Running ncn-healthcheck-worker.yaml...DONE
Running ncn-healthcheck-worker-single.yaml...DONE
Running ncn-afterpitreboot-healthcheck-worker.yaml...DONE
Running ncn-afterpitreboot-healthcheck-worker-single.yaml...DONE
Running ncn-kubernetes-tests-worker.yaml...DONE
Running ncn-afterpitreboot-kubernetes-tests-worker-single.yaml...DONE
Running ncn-cms-tests.yaml...DONE
Aggregated goss results have been saved to: /tmp/tmp.pLpgqaGT8V
DST-and-ct-results-compatible file been saved to: /tmp/api-results.json
ncn-w002:~/ # jq '.tests | length' < /tmp/api-results.json
119
```

**Storage**

```
ncn-s001:~ # ./dst-ct-results.sh
Found 6 suites to run
Running ncn-preflight-tests.yaml...DONE
Running ncn-smoke-tests.yaml...DONE
Running ncn-spire-healthchecks.yaml...DONE
Running ncn-healthcheck-storage.yaml...DONE
Running ncn-afterpitreboot-healthcheck-storage.yaml...DONE
Running ncn-storage-tests.yaml...DONE
Aggregated goss results have been saved to: /tmp/tmp.8UfELQmAS1
DST-and-ct-results-compatible file been saved to: /tmp/api-results.json
ncn-s001:~ # jq '.tests | length' < /tmp/api-results.json
96
```

I also submitted the resulting files to the dashboard and saw they
reported appropriately.

```
HTTP/2 200
server: nginx/1.24.0
date: Wed, 16 Oct 2024 19:29:54 GMT
content-type: application/json
content-length: 341
access-control-allow-origin: *

{
  "artifacts": {
    "github_url": "",
    "run_id": "2234567890",
    "run_url": "
  },
  "success": true,
  "trace_id": "",
  "warnings":[]
}
```

### DIFF
--- a/goss-testing/automated/dst-ct-results.sh
+++ b/goss-testing/automated/dst-ct-results.sh
@@ -1,0 +1,380 @@
+#!/usr/bin/env bash
+#
+# MIT License
+#
+# (C) Copyright 2024 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# Run CSM goss tests and produce a DST-compatible results.json file.
+# 
+# 1. Query the cgroups file for the pids of the goss suites/vars that are linked to an http endpoint
+# 2. Get the command each PID is running 
+# 3. Reformat the command into a goss command that can be run directly instead of with 'goss serve'
+# 4. Loop through and run each newly-formatted goss command and save the results to a file
+# 5. Aggregate each iteration into an api-results.json file, used by DST (additional formatting is required)
+# 6. Save the file, which can then be slurped up by the dst pipeline and sent to the results dashboard
+set -euo pipefail
+
+#######################################
+# Check for required RPMs, commands, and other prerequisites.
+# Globals:
+#   None
+# Arguments:
+#   ostype: the operating system type (default is $OSTYPE)
+# Outputs:
+#   None
+# Returns:
+#   0 if reqs are met, non-zero if not.
+#######################################
+prereqs() {
+  local rc=0
+  local reqs_rpms reqs_cmds
+  local ostype="${1:-${OSTYPE}}"
+
+  if [[ "${ostype}" != "linux"* ]]; then
+    echo "This script is only supported on Linux, detected: $ostype" >&2
+    return 1
+  fi
+
+  reqs_rpms=(
+    csm-testing 
+    goss-servers
+  )
+  for req in "${reqs_rpms[@]}"; do
+    if ! rpm -q "$req" &>/dev/null; then
+      echo "Required RPM is not installed: $req" >&2
+      rc=1
+    fi
+  done
+
+  reqs_cmds=(
+    jq
+    kubectl
+    yq
+  )
+  for req in "${reqs_cmds[@]}"; do
+    if ! command -v "$req" &>/dev/null; then
+      echo "Required command is not available: $req" >&2
+      rc=1
+    fi
+  done
+
+  if [[ "${rc:-0}" -ne 0 ]]; then
+    echo "Please address the missing prerequisites and try again." >&2
+    return $rc
+  fi
+
+  return 0
+}
+
+usage() {
+  # echo the usage in order to use the variable for the script name
+  # everything else is in the comments
+  echo "Usage: $(basename -- "${0}") [-h] OUTPUT_FILE"
+  # Any line startng with with a #/ will show up in the usage line
+
+  #/
+  #/    Run CSM goss tests and produce a DST-compatible results.json file.
+  #/
+  #/    -h      Display this help message
+  #/
+  grep '^  #/' "$0" | cut -c6-
+  return 0
+}
+
+#######################################
+# Set global variables for the script
+# Globals:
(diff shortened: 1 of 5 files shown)


goss-testing/automated/dst-ct-results.sh
```diff
@@ -0,0 +1,379 @@
+#!/usr/bin/env bash
+# MIT License
+#
+# (C) Copyright 2023-2024 Hewlett Packard Enterprise Development LP
+#
+# Permissioff is hereby granted, free of charge, to any persoff obtaining a
```
shunr-hpe: Permissioff typo and other typos: persoff, documentatioff, ACTIoff, CONNECTIoff

spec/functional/dst-ct-results_spec.sh
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env bash
+# MIT License
+#
+# (C) Copyright 2023-2024 Hewlett Packard Enterprise Development LP
+#
+# Permissioff is hereby granted, free of charge, to any persoff obtaining a
```
shunr-hpe, line 6: typo

shunr-hpe (approved):

PR #607, opened 2024-10-21
use find cmd instead of static path for finding goss-server pids

jacobsalmela:
the path the file is found at can vary between nodes, so using the find command is safer and more resilient.


dborman-hpe (approved):
LGTM

PR #611, opened 2024-10-28
quit find after one match, adjust test accordingly

jacobsalmela:


dborman-hpe (approved):
LGTM.
FYI I checked this on mug, and verified that it doesn't matter which of the multiple responses is used, they all report the same information.
```
ncn-m001:~ # md5sum `find /sys/fs/cgroup -path '*/goss-servers.service/cgroup.procs' -type f`
21ca286f5674493404c310e736b3a70d  /sys/fs/cgroup/pids/system.slice/goss-servers.service/cgroup.procs
21ca286f5674493404c310e736b3a70d  /sys/fs/cgroup/blkio/system.slice/goss-servers.service/cgroup.procs
21ca286f5674493404c310e736b3a70d  /sys/fs/cgroup/memory/system.slice/goss-servers.service/cgroup.procs
21ca286f5674493404c310e736b3a70d  /sys/fs/cgroup/devices/system.slice/goss-servers.service/cgroup.procs
21ca286f5674493404c310e736b3a70d  /sys/fs/cgroup/cpu,cpuacct/system.slice/goss-servers.service/cgroup.procs
21ca286f5674493404c310e736b3a70d  /sys/fs/cgroup/systemd/system.slice/goss-servers.service/cgroup.procs
21ca286f5674493404c310e736b3a70d  /sys/fs/cgroup/unified/system.slice/goss-servers.service/cgroup.procs
```

jacobsalmela:
Yeah @dborman-hpe I noticed this as well, which is why I changed it to just one match.  There may be a better way to get the pids with `pgrep` or some other method, but I usually start with the most basic tools and work up from there.  This one just needed several refinements after working it across many different systems. 

Thank you!

PR #615, opened 2024-10-29
CASMINST-6936 update readme for running goss tests locally

jacobsalmela:


goss-testing/README.md
```diff
@@ -88,6 +88,27 @@ Use the following command to run a single test or a suite of tests on the local
...
 ```
 
+## Running All Test Suites Locally
+
+The goss tests can be executed directly (not via the http endpoint) on any node type using a single script.  This will save both a GOSS-formatted JSON file and a DST-compatible JSON file.  
+
+It will run any test that is currently part of a suite connected to a GOSS endpoint.
+
+```bash
+/opt/cray/tests/install/ncn/automated/dst-ct-results.sh
```
mharding-hpe, line 100: The paragraph above seems to indicate you can choose which suite you want it to run. Can you provide a usage example of how to specify that? Do you provide the full path to the suite?

Oh, or are you saying that it will run EVERY test suite that is associated with a GOSS endpoint? I assume it also filters based on the current node type?

Maybe clarify this slightly in the preceding test.

mharding-hpe (approved):

goss-testing/README.md
jacobsalmela: Sorry @mharding-hpe for the delayed response.  Yes, it runs every suite currently for use in the CT pipeline/nightly testing.  I will update the wording.

PR #617, opened 2024-10-31
CASMINST-6951: Move Python code into virtual environment

mharding-hpe:
This PR moves the `csm-testing` Python code into its own Python package, and then installs this package into a dedicated `csm-testing` Python environment. The Python code itself is essentially unchanged (except for minimal changes required to accomplish the above -- no functional changes).

This involved a fair amount of work inside the RPM spec file, and while I was doing it, I noticed that our tests and tools rely on a good number of packages that are not marked as requirements of the RPM. So I also updated those.

Functionally, all of the above changes should be transparent to the end user. I also tried to do this work with an eye toward making it easy to add new Python tests or tools, without having to make changes in a bunch of different places. In particular, just adding a new test or tool into the right place in the `csm-testing` Python package will automatically generate a shortcut link to that test or tool inside of the RPM install. See the updated README file for details on that.

When adding Python tests that require additional packages, they can be added to the usual Python `requirements.txt` and `constraints.txt` files, and they will be installed into the virtual environment automatically.

It is not part of this PR, but I also want to incorporate pylint and (eventually) mypy into the build process, to help us catch code errors earlier. But doing that will be a significant effort because of the problems that will no doubt be found in the legacy Python code. So I will leave that for future PRs.

In addition, this PR uses Python 3.6 for the virtual environment (although I tried to write everything in such a way that we can change this to a newer version without having to make changes in a ton of places). I had originally planned to use Python 3.11, since that is available on all NCNs, but I ran into trouble. Specifically, one of the Python tests uses the `rados` Python package. This package is not published to Pypi, nor to our internal CSM Python repos. As far as I can tell, it's only available as a Python 3.6 RPM package. There is no clean way to install that into a virtual environment, much less one for a newer Python version. Because of this, the PR creates a virtual environment which inherits some packages from the system Python environment. Basically, if a dependency is not found in the virtual environment, it will then go look in the system Python environment.

If we remove this dependency on the `rados` package, then we could remove the inheritance from the system Python environment, and also move to a newer Python version. But that effort can be saved for a future ticket.

I have tested this on mug and found no problems caused by the changeover.


mharding-hpe:
It occurs to me that only a single test (`check_for_unused_drives`) uses the `rados` package. Another option would be to allow that single Python test to exist outside of the virtual environment. It could use the system Python (like we currently do for everything), and then everything else could go into the self-contained virtual environment, with nothing inherited from the system. But I can do this with a follow-up ticket, unless we really want it in this one.

I'd still want to do further testing before moving up from Python 3.6, though. IMO that should be reserved for CSM 1.7.

mharding-hpe:
When this merges, I plan to tag it with a bumped minor number. That way we will still be able to make fixes (if needed) for CSM 1.6.0 without pulling this in

jacobsalmela:
Still looking.  I want to try it out on a machine, which will likely be tomorrow

mharding-hpe:
> Still looking.  I want to try it out on a machine, which will likely be tomorrow

No worries -- no rush from my end

jacobsalmela (approved):

PR #618, opened 2024-11-01
CASMPET-7266: Fix hostname regex in start-goss-servers.sh

mharding-hpe:
On the CSM 1.6.1 vshasta deployment last night, the `goss-servers` service failed to start properly on the PIT node, because of a bad regex in my PR for [CASMPET-7263](https://jira-pro.it.hpe.com:8443/browse/CASMPET-7263).

This PR corrects the regex, making it exactly match the ones used in `lib/endpoints.py`, which is what the `goss-servers` service uses to determine what test endpoints to start up.

I also added a more detailed comment about the hostname check and the corresponding regex, making the link between those two things clear.

There are backports for CSM 1.5 and CSM 1.4, since this change went into those as well:
1.5: https://github.com/Cray-HPE/csm-testing/pull/619
1.4: https://github.com/Cray-HPE/csm-testing/pull/620

### DIFF
--- a/systemd/start-goss-servers.sh
+++ b/systemd/start-goss-servers.sh
@@ -34,10 +34,18 @@ fi
 # necessary for kubectl commands to run
 export KUBECONFIG=/etc/kubernetes/admin.conf
 
-# necessary for test that need to know the current hostname
+# Some tests need to know the current hostname.
+#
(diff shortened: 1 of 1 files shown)


bo-quan (approved):

dborman-hpe (approved):
LGTM

PR #624, opened 2024-11-05
CASMTRIAGE-7457: Fix bug causing some Python script symlinks from not being created

mharding-hpe:
An error in the regular expressions used in the `create-python-script-symlinks.sh` script resulted in no symlinks being created for some Python scripts in the virtual environment. Specifically, any scripts whose names contained numbers were omitted.

This PR refactors the build shell scripts to remove the need for that kind of regular expression. Now the script that creates the symlinks does not need to parse and transform the output from the script which updates pyproject.yaml. Instead, that latter script now adds two different types of lines to pyproject.yaml. One is the same as before, to be consumed by Python setuptools when building the package. The second is a comment line which provides exactly the data needed as input for the script that creates the Python symlinks. 

This hopefully reduces the chances of a bug like the one we saw, where the script was registered with the Python module, but no symlink was created.

I have tested this on mug and verified that the missing symlinks are now created.

This PR also adds me and @jacobsalmela to the codeowners file, since we do a lot of work in the repo, and probably should get tagged as PR reviewers.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,4 @@
 *       @Cray-HPE/metal
 *       @Cray-HPE/platform-engineering
-
+*       @mharding-hpe
+*       @jacobsalmela
--- a/common.sh
+++ b/common.sh
@@ -22,7 +22,15 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 
+# No shebang line included in this file because it is only intended to be sourced, never
+# directly executed. But to let shellcheck know:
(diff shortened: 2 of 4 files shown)


jacobsalmela (approved):
I didn't extensively test #617, just some quick manual smoke tests 🍳

mharding-hpe:
> I didn't extensively test #617, just some quick manual smoke tests 🍳

I ran a fair amount of tests, but I guess I missed running the one test which used the one script whose symlink wasn't being created. :)

PR #638, opened 2024-12-05
CASMPET-7290: Add pylint to build pipeline; correct errors identified by pylint

mharding-hpe:
This PR adds pylint to the build pipeline. The builds will only fail if pylint identifies fatal errors. A full pylint run will also happen, but currently it will not be able to cause the build to fail. This is because there are a lot of issues it identified in the existing Python code. Hopefully we can gradually address those. If so, we can eventually have pylint require minimum code quality, or else fail the build.

After adding pylint, it identified several fatal errors in the existing tests. This PR corrects those as well. None of them are tricky or risky changes.

### DIFF
--- a/Jenkinsfile.github
+++ b/Jenkinsfile.github
@@ -82,6 +82,22 @@ pipeline {
                 sh "make pymod"
             }
         }
+        stage('Validate Python module') {
+            agent {
+                docker {
+                    label "${PRIMARY_NODE}"
+                    reuseNode true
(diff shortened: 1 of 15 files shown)


jacobsalmela (approved):
Cool thanks

mharding-hpe:
> Cool thanks

No problem -- this is something I've been trying to do in a lot of repos. I also want to eventually add mypy, but for that we'll need to make sure the code has type annotations, and that'll also take a little while. But that has allowed me to find a bunch of bugs in BOS, so I'm hoping to eventually get this done.

PR #652, opened 2025-01-07
CASMINST-7108 Simplify filename pattern override for license checker

mtupitsyn:
## Summary and Scope

New version of license checker supports filename pattern config as `<pattern> => {type: <type}` map. Change is backwards compatible, old config format is also recognized by license checker. However, new format allows overriding for individual entries, instead of replicating full list. 

## Issues and Related PRs

* Resolves [CASMINST-7108](https://jira-pro.it.hpe.com:8443/browse/CASMINST-7108)

### Tested on:

### Test description:
    
    # test pattern defined in override
    $ docker run -it --rm -v $(pwd):/github/workspace artifactory.algol60.net/csm-docker/stable/license-checker --log-level=debug csm-testing.spec | grep matched
    [DEBUG] Matching csm-testing.spec against *.spec to determine file type .... matched!
    
    # test pattern defined in base config
    $ docker run -it --rm -v $(pwd):/github/workspace artifactory.algol60.net/csm-docker/stable/license-checker --log-level=debug goss-testing/scripts/ceph-service-status.sh | grep matched
    [DEBUG] Matching goss-testing/scripts/ceph-service-status.sh against *.sh to determine file type .... matched!

## Risks and Mitigations

Low - build time check only.

### DIFF
--- a/.license_check.yaml
+++ b/.license_check.yaml
@@ -23,12 +23,14 @@
 #
 
 # Define custom types for our script files which should have licenses
-# but do not match a default pattern. Use yaml for .spec and .service
+# but do not match a default pattern. Use yaml for .spec, .service and .preset
 # files -- they are not in YAML, but as far as license checking is
 # concerned, they can be considered the same.
 
 file_types:
   "*.service":
+    type: yaml
+  "*.preset":
     type: yaml
   "*.spec":
     type: yaml
@@ -36,5 +38,3 @@ file_types:
     type: shell_or_python
   "build-testing/*":
     type: shell_or_python
-
-
--- a/.license_check.yaml
+++ b/.license_check.yaml
@@ -1,7 +1,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2022 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022,2025 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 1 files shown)


.license_check.yaml
```diff
@@ -28,51 +28,13 @@
...
-  - pattern: "*.bash"
-    type: shell_or_python
-  - pattern: "*.py"
-    type: shell_or_python
-  - pattern: "Dockerfile*"
-    type: docker
-  - pattern: "*/Dockerfile*"
-    type: docker
-  - pattern: "*.java"
-    type: java
```
mtupitsyn, line 46: Override is already deviated from base config here: `java` is actually split into `java_inline` and `java_block`.

.license_check.yaml
```diff
@@ -28,51 +28,13 @@
...
 
 file_types:
-  - pattern: "*.service"
+  "*.service":
     type: yaml
-  - pattern: "*.spec"
+  "*.spec":
     type: yaml
-  - pattern: "*.sh"
+  "goss-testing/automated/*":
```
jacobsalmela, line 37: There are also scripts in:

- `goss-testing/scripts`
- `systemd`

Those have licenses already so should also be included

jacobsalmela (changes requested):

.license_check.yaml
mharding-hpe: Also under `src`

mharding-hpe (changes requested):
Should include the `src` directory when checking for Python script license headers

.license_check.yaml
mtupitsyn: I see that all files under `goss-testing/scripts/` have proper `.sh` extension, so they are covered by base default configuration, which associates `*.sh` with type `shell_or_python`. Same for file under `src/` - they all match `*.py` and thus treated as Python files. For files under `systemd/`, `goss-servers.service` is covered by `*.service` pattern defined above, and `start-goss-servers.sh` is a regular shell script covered by base config. So the only file not covered is `systemd/90-goss-servers.preset`. I'll add `systemd/*.preset` pattern.

mtupitsyn: Done. FYI, there's few files not recognized by base config, and few files missing license header:

    $ docker run -it --rm -v $(pwd):/github/workspace artifactory.algol60.net/csm-docker/stable/license-checker | grep -E 'not (detected|recognized)'
    [INFO] Filename pattern not recognized: ./pyproject.toml
    [INFO] Filename pattern not recognized: ./.shellspec
    [INFO] Filename pattern not recognized: ./goss-testing/dat/goss-servers.cfg
    [WARNING] License is not detected: ./spec/functional/dst-ct-results_spec.sh
    [WARNING] License is not detected: ./.github/dependabot.yml
    [WARNING] License is not detected: ./.github/workflows/merge-strategy.yml

Unrecognized file types are treated as INFO level (non concerning) by license checker and can be ignored (unless you specifically want these files to have license). Missing licenses in recognized file types are reported by license checker as check failure.

mharding-hpe (approved):

jacobsalmela (approved):

PR #655, opened 2025-01-07
CASMTRIAGE-7627 ignore non-csm failed backups

jacobsalmela:
`velero_backups_check.sh` causes a test failure intermittently on in-house sytems, often from other products such as PBS or Slingshot. While they may be valid failures, the CSM product is not at fault, so those should not trigger a test failure.

for this commit, I modified it to check only for failed vault backups since that is more pertinent to CSM itself.

I moved the code that was previously run with a pipe and checking the exit code to instead make a temp file, use a single jq command to query it, and then fail if there are any failed vault backups.

if we determine other backups should be checked, we can add those to
    that jq command.

### DIFF
--- a/goss-testing/scripts/velero_backups_check.sh
+++ b/goss-testing/scripts/velero_backups_check.sh
@@ -24,7 +24,7 @@
 #
 
 # Due to a velero bug, a backup is created anytime the backup schedule is created or updated.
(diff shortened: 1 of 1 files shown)


mharding-hpe (approved):

kimjensen-hpe (approved):
Consider changing the variable “failed_bakup_path” to “failed_backups_path”

spillerc-hpe (approved):

PR #659, opened 2025-01-08
CASMTRIAGE-7657 adjust k8s_verify_cluster_2 test to avoid some false

jacobsalmela:
positives by checking if a newer pod was successful.

the test currently fails when any output is detected.  it does so when any error pod is detected.  however, there are cases when one pod fails, but another is started and is then successful.

this commit adds extra logic to check if a newer pod exists and is a non-error state, which is then presumed to be acceptable.  no output is printed in this case.

otherwise, the podname is printed triggering the test to fail.

it was necessary to move the code to an external script because the
logic and special characters were proving too cumbersome to handle in
the limited `exec:` block

### DIFF
--- a/goss-testing/scripts/verify_kube_system_pods.sh
+++ b/goss-testing/scripts/verify_kube_system_pods.sh
@@ -2,7 +2,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2022-2024 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2022-2025 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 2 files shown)


goss-testing/scripts/verify_kube_system_pods.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+#
+# MIT License
+#
+# (C) Copyright 2022-2024 Hewlett Packard Enterprise Development LP
```
mharding-hpe:
```suggestion
# (C) Copyright 2022-2025 Hewlett Packard Enterprise Development LP
```

mharding-hpe (approved):

PR #674, opened 2025-01-31
CASMNET-2204

marumugam-hpe:
This test ensures that if cilium is installed in a system, it checks the running phase count. In case if weave is present instead of cilium, it runs a check for weave status. 

JIRA: https://jira-pro.it.hpe.com:8443/browse/CASMNET-2204

This was tested on Wasp, since it has cilium installed.


goss-testing/scripts/check_cilium.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+#
+# MIT License
+#
+# (C) Copyright 2024 Hewlett Packard Enterprise Development LP
```
rustydb:
```suggestion
# (C) Copyright 2025 Hewlett Packard Enterprise Development LP
```

goss-testing/tests/ncn/goss-cilium-health.yaml
```diff
@@ -0,0 +1,39 @@
+# (C) Copyright 2014-2023 Hewlett Packard Enterprise Development LP
```
rustydb:
```suggestion
# (C) Copyright 2025 Hewlett Packard Enterprise Development LP
```

goss-testing/scripts/check_cilium.sh
```diff
@@ -0,0 +1,37 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# This test checks that if cilium is up and running in case cilium is installed
+
```
rustydb:
```suggestion
set -euo pipefail
```
May want to retest the script after this change.

goss-testing/scripts/check_cilium.sh
```diff
@@ -0,0 +1,37 @@
...
+if [ "$(kubectl get pods -n kube-system -o wide | grep -E 'cilium')" ]; then
```
rustydb: Set and check a variable instead.
e.g.
```bash
foo="$(kubectl get pods -n kube-system -o wide | grep -E 'cilium')"
if [ -z "$foo" ];
```
or, check the return code from `grep` by using the `-q` flag.
```bash
if kubectl get pods -n kube-system -o wide | grep -qE 'cilium' ; then
```

goss-testing/scripts/check_cilium.sh
```diff
@@ -0,0 +1,37 @@
...
+  phase_count=$(cilium status -o json | jq -r '.phase_count.cilium.Running')
+  if [ "$phase_count" -gt 0 ]; then
+    echo "PASS"
+  fi
+else
+  if [ "$(weave --local status connections)" ]; then
```
rustydb: Does `weave --local status connections` ever output nothing?

jacobsalmela: I think if you are just checking the exit status, you can simplify this:
```suggestion
  if weave --local status connections; then
```

goss-testing/scripts/check_cilium.sh
jacobsalmela: This command doesn't necessarily mean the pod is running.  Also getting it by label is probably better in the case several pods are needed.  This also reduces the pipes, which aren't inherently bad, but it is nice to use one command when you can.

```suggestion
kubectl get pods -n kube-system -l name=cilium -o jsonpath='{.spec.state}'
# check output if it is running or other desired state
```

jacobsalmela (changes requested):

rustydb (approved):

rustydb (approved):

jacobsalmela (approved):

jpdavis-prof (approved):

PR #693, opened 2025-03-21
CASMPET-7103, CASMTRIAGE-7780 fix intermittent failures in `check_key_id_in_jwks.sh`

jacobsalmela:
autotriage has detected intermittent, but continued failures with this script.  I have seen failures at all exit codes, `10`, `30`, and `50`.

first, this reduces pipes when calling `heartbeat-spire-agent api fetch jwt` by using `-output json` and then parsing that output with `jq`. this works more efficiently by ensuring the output is machine-readable. this should address exit `10` and make it more resilient.

after the web token is in the `$jwt` variable, I extract the header, leaving room for expansion of this script for the other two parts, payload and signature.  per the munging happening to make the base64 decoder happy, there are a few more cases the current code did not account for.  I mentioned those in a comment and I changed the decode logic to be in its own function and then decode the header passed as input to that function.

exit `30` has reduced pipes now because the decoding is done prior to that, and it is simply a small `jq` call now.

exit `50` should be more resilient by way of the other two changes.  I also added `-s` to curl to reduce noisy output.  this helps when testing the script in parallel at scale and parsing for pass/fail.

### DIFF
--- a/goss-testing/scripts/check_key_id_in_jwks.sh
+++ b/goss-testing/scripts/check_key_id_in_jwks.sh
@@ -25,26 +25,45 @@
 
 set -euo pipefail
 
-jwt=$(/usr/bin/heartbeat-spire-agent api fetch jwt -socketPath /var/lib/spire/agent.sock -audience goss-test \
-  | head -n2 | awk -F. 'FNR==2{sub(/^[ \t]+/, ""); print $1}') || exit 10
+jwt=$(/usr/bin/heartbeat-spire-agent api fetch jwt -socketPath /var/lib/spire/agent.sock -audience goss-test -output json \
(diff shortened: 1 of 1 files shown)


ndavidson-hpe (approved):

goss-testing/scripts/check_key_id_in_jwks.sh
mharding-hpe, line 1: Need to bump the copyright year in the header

mharding-hpe (approved):

goss-testing/scripts/check_key_id_in_jwks.sh
jacobsalmela: Fixed.

PR #700, opened 2025-04-10
Casmtriage 8042 casmpet 7481 fix

aasha-hpe:
## Summary and Scope

GOSS tests for iSCSI SBPS is the preliminary test run to check the health of the node related to iSCSI SBPS. The fix now is to report errors for the failure scenarios so that appropriate action can be taken. 

## Issues and Related PRs
Related issues  CASMTRIAGE-7976 where NMN A records were not present and iSCSI SBPS boot via NMN failed. 

* Resolves [issue id](issue link): CASMTRIAGE-8042 and CASMPET-7481
* Change will also be needed in `<insert branch name here>`: N/A 
* Future work required by [issue id](issue link): N/A 
* Documentation changes required in [issue id](issue link): N/A 
* Merge with/before/after `<insert PR URL here>`: N/A

## Testing
Tested on Tyr where ncn health checks were run on worker nodes and observed that goss tests were run and tests have passed.

### Tested on:
Tyr

### Test description:

The goss test script "iscsi_sbps_sanity.sh" with the fix was copied onto worker nodes under "/opt/cray/tests/install/ncn/scripts/" directory and /opt/cray/tests/install/ncn/automated/ncn-healthcheck-worker ran from master node (m001). Verified the logs on the worker nodes under /opt/cray/tests/install/logs/iscsi_cps_sanity

- Were the install/upgrade-based validation checks/tests run (goss tests/install-validation doc)?: ?This goss test itself
- Were continuous integration tests run? If not, why?: N/A 
- Was upgrade tested? If not, why?: N/A 
- Was downgrade tested? If not, why?: N/A 
- Were new tests (or test issues/Jiras) created for this change?: N/A 

## Risks and Mitigations
None

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [NA] CHANGELOG.md updated
- [NA] [HPC Product Announcement](https://cray.slack.com/archives/C026TVCSXLH) prepared, if applicable

### DIFF
--- a/goss-testing/scripts/iscsi_sbps_sanity.sh
+++ b/goss-testing/scripts/iscsi_sbps_sanity.sh
@@ -2,7 +2,7 @@
 #
 # MIT License
 #
-# (C) Copyright 2024 Hewlett Packard Enterprise Development LP
+# (C) Copyright 2025 Hewlett Packard Enterprise Development LP
 #
 # Permission is hereby granted, free of charge, to any person obtaining a
 # copy of this software and associated documentation files (the "Software"),
@@ -25,8 +25,6 @@
 # This script is to do some sanity tests for iSCSI based boot content 
 # projection.
 
-# shellcheck disable=SC2207,SC2143,SC2002
-
 LIO_ACT=0
 
 echo "Checking whether the node is configured with iSCSI"
@@ -41,10 +39,19 @@ else
     echo "${HOSTNAME} is configured with iSCSI"
 fi
 
-# Verifying whether Marshal agent is running or not
+EXIT_VAL=0
+
+# Verifying whether Marshal agent is running or not and if there are any errors
 
 echo "*******************************"
 echo "Marshal agent is $(systemctl is-active sbps-marshal.service)"
+
+ERR=$(journalctl -xeu sbps-marshal.service | tac | sed '/START SCAN/q' | grep -i error)
+
+if [ -n "${ERR}" ]; then
+    echo "Marshal agent is having errors, please check"
+    EXIT_VAL=1
+fi
 echo "*******************************"
 
 # Verifying whether LIO target(s) is active
@@ -62,6 +69,7 @@ echo "*******************************"
 
 # Verify TCP service probes complete against all active portals (iSCSI)
 
+# shellcheck disable=SC2207
 PORTALS=($(ss -tnpl | grep ':3260 ' | awk '{ print $4 }' ))
 
 if [[ -n "${PORTALS[*]}" ]]; then
@@ -104,9 +112,11 @@ if [ -s tmp_file ];then
     echo "*******************************"
 else
     echo "DNS SRV records do not exist, please create them"
+    rm tmp_file
     exit 1
 fi
 
+# shellcheck disable=SC2207,SC2002
 SRV=($(cat tmp_file | awk '{print $(NF)}' | sed -e 's/\.$//'))
 
 echo "DNS A records are as below"
@@ -114,7 +124,13 @@ echo "--------------------------"
 
 for i in "${SRV[@]}"
 do
-    dig -t A +short "${i}"
+    DNS_A=$(dig -t A +short "${i}")
+    if [ -z "${DNS_A}" ]; then
+        echo "ERROR: DNS A record is absent for ${i}"
+        EXIT_VAL=1
+    else
+        echo "DNS A record is present: ${DNS_A} for ${i}"
+    fi
 done
 
 rm tmp_file
@@ -128,3 +144,5 @@ echo "*********************************************"
 for s in $(dig -t srv +short _sbps-hsn._tcp."${host}" | sort -k3 | awk '{print $NF;}' | xargs); do printf '** %s **\n' "$s"; dig +short "$s"; done
 
 for s in $(dig -t srv +short _sbps-nmn._tcp."${host}" | sort -k3 | awk '{print $NF;}' | xargs); do printf '** %s **\n' "$s"; dig +short "$s"; done
+
+exit ${EXIT_VAL}


goss-testing/scripts/iscsi_sbps_sanity.sh
```diff
@@ -114,7 +120,12 @@ echo "--------------------------"
 
 for i in "${SRV[@]}"
 do
-    dig -t A +short "${i}"
+    DNS_A=$(dig -t A +short "${i}")
+    if [ -z "${DNS_A}" ]; then
+        echo "ERROR: DNS A record is absent for ${i}"
```
dborman-hpe, line 129: Should this cause a non-zero exit status?

goss-testing/scripts/iscsi_sbps_sanity.sh
```diff
@@ -41,10 +41,16 @@ else
...
-# Verifying whether Marshal agent is running or not
+# Verifying whether Marshal agent is running or not and if any errors
 
 echo "*******************************"
 echo "Marshal agent is $(systemctl is-active sbps-marshal.service)"
+
+ERR=$(journalctl -xeu sbps-marshal.service | grep -i error)
+
+if [ ! -z "${ERR}" ]; then
+    echo "Marshal agent is having errors, please check"
```
dborman-hpe, line 52: Should this cause a non-zero exit status?  But it's a very broad check so if there was ever an error, this would always say there are errors, so exiting non-zero would probably mean once there was an error, this script would always fail, which wouldn't be good.

goss-testing/scripts/iscsi_sbps_sanity.sh
```diff
@@ -114,7 +120,12 @@ echo "--------------------------"
...
-    dig -t A +short "${i}"
+    DNS_A=$(dig -t A +short "${i}")
+    if [ -z "${DNS_A}" ]; then
+        echo "ERROR: DNS A record is absent for ${i}"
+    else
+        echo "DNS A record is present: ${DNS_A} for ${i}"
+    fi
 done
 
 rm tmp_file
```
dborman-hpe, line 136: There should probably be an explicit exit at the end, rather than returning the value from the `rm tmp_file` command.  Also, up above at line 107/113 there is an `exit 1` that will leave `tmp_file` lying around.

goss-testing/scripts/iscsi_sbps_sanity.sh
aasha-hpe: Yes, I was thinking NMN A record being absent where we can still boot over HSN where we log a message (not to exit) as its very unlikely to hit HSN A record being absent. What do you suggest ?

goss-testing/scripts/iscsi_sbps_sanity.sh
aasha-hpe: Yes and also if Marshal agent is having errors better to report it and execute rest of the goss script where it reports about active portals, DNS SRV & A records etc.. Instead of journalctl, should we check for "systemctl status sbps-marshal.service  | grep- i error" to look for current errors ?

goss-testing/scripts/iscsi_sbps_sanity.sh
dborman-hpe: There are 3 situations:

1. we have all the A records
2. we have some of the A records
3. we have none of the A records

The first case is success, the last case is failure.  The issue we had before was the NMN records weren't getting generated, but nothing flagged that, so is this the place to enforce that?  But the missing NMN records was due to a coding error that we've fixed.

This loop is going through the results from looking for
```
_sbps-hsn._tcp."${host}" _sbps-nmn._tcp."${host}" 
```
entries, so since we've set them up, we should have set up A records for each.  So I'm thinking any missing A records should cause a failure.

It doesn't have to exit at this point, just set a flag for a non-zero exit so it can do the rest of the checks and then exit non-zero at the end.

goss-testing/scripts/iscsi_sbps_sanity.sh
dborman-hpe: So maybe the thing to do is to look for errors in just the latest scan, i.e. the output between the last `START SCAN` and `END SCAN`.  After a little playing around with `sed` and doing some google searches, this should work:
```
journalctl -xeu sbps-marshal.service | tac | sed '/START SCAN/q' | grep -i error
```
The `tac` call reverses the input, the `sed` command quit once it finds the first `START SCAN`, which is really the last `START SCAN` to the end, since the input was reversed.

And instead of exiting if errors are found, set a flag that the script should exit non-zero so that the rest of the tests can be run.

goss-testing/scripts/iscsi_sbps_sanity.sh
aasha-hpe: yes, this is the place to flag that A records are not generated. Your suggestion to set a flag and a non-zero exit at the end sounds good to me.

goss-testing/scripts/iscsi_sbps_sanity.sh
aasha-hpe: Any reason for not chosing "systemctl status sbps-marshal.service | grep -i error"  which I believe report latest scan status ?.

goss-testing/scripts/iscsi_sbps_sanity.sh
```diff
@@ -41,10 +41,16 @@ else
     echo "${HOSTNAME} is configured with iSCSI"
 fi
 
-# Verifying whether Marshal agent is running or not
+# Verifying whether Marshal agent is running or not and if any errors
```
dborman-hpe: How about
`# Verify whether or not the Marshal agent is running, and if there are any errors

goss-testing/scripts/iscsi_sbps_sanity.sh
dborman-hpe: The `systemctl status` only reports the last few lines from the log, not the entire last scan.  On gamora the scans are generating 330 lines per scan. Hmm... I noticed that there are only 1000 lines coming out, not the entire history.  That's due to the `-e` flag:
```
ncn-w001:~ # journalctl -u sbps-marshal.service | wc -l
37488
ncn-w001:~ # journalctl -eu sbps-marshal.service | wc -l
1000
```
So the `-e` is already limiting it to the last 1000 lines, my command above further limits it to just the last scan.

aasha-hpe: okay

goss-testing/scripts/iscsi_sbps_sanity.sh
aasha-hpe: Yes,  looks good.

goss-testing/scripts/iscsi_sbps_sanity.sh
aasha-hpe: I have tried to address this, Pls see the latest changes, hope this is what you meant. pls let me know otherwise.

goss-testing/scripts/iscsi_sbps_sanity.sh
```diff
@@ -128,3 +143,9 @@ echo "*********************************************"
 for s in $(dig -t srv +short _sbps-hsn._tcp."${host}" | sort -k3 | awk '{print $NF;}' | xargs); do printf '** %s **\n' "$s"; dig +short "$s"; done
 
 for s in $(dig -t srv +short _sbps-nmn._tcp."${host}" | sort -k3 | awk '{print $NF;}' | xargs); do printf '** %s **\n' "$s"; dig +short "$s"; done
+
+if [ -z "$A_REC" ]; then
```
dborman-hpe: This is testing if the string `A_REC` is is empty, which will always evaluate to false (it's either `1` or `0`).  Instead of `A_REC`, at the top of the script set `EXIT_VAL=0`, then when an A record is missing set `EXIT_VAL=1`.  Then there is no need to check anything here, just end with `exit ${EXIT_VAL}`

goss-testing/scripts/iscsi_sbps_sanity.sh
dborman-hpe: We should now also set `EXIT_VAL=1`, since we have limited the error checking to just the last scan.  See my comments about the A records an using `EXIT_VAL`.

goss-testing/scripts/iscsi_sbps_sanity.sh
```diff
@@ -41,10 +41,16 @@ else
     echo "${HOSTNAME} is configured with iSCSI"
 fi
```
dborman-hpe, line 41: About here is where to add
```
EXIT_VAL=0
```

goss-testing/scripts/iscsi_sbps_sanity.sh
dborman-hpe: Change the line below from `A_REC=0` to `EXIT_VAL=1`

goss-testing/scripts/iscsi_sbps_sanity.sh
```diff
@@ -104,17 +110,26 @@ if [ -s tmp_file ];then
...
 else
     echo "DNS SRV records do not exist, please create them"
+    rm tmp_file
     exit 1
 fi
 
 SRV=($(cat tmp_file | awk '{print $(NF)}' | sed -e 's/\.$//'))
 
+A_REC=1
+
```
dborman-hpe: This can be removed when switching to setting `EXIT_VAL` up above.

dborman-hpe:
Here's a summary of the final changes I'm requesting.  This includes changing the shellcheck disable from global to the file, to on the specific lines where they are needed.
```
--- a/goss-testing/scripts/iscsi_sbps_sanity.sh
+++ b/goss-testing/scripts/iscsi_sbps_sanity.sh
@@ -25,8 +25,6 @@
 # This script is to do some sanity tests for iSCSI based boot content 
 # projection.
 
-# shellcheck disable=SC2207,SC2143,SC2002
-
 LIO_ACT=0
 
 echo "Checking whether the node is configured with iSCSI"
@@ -41,6 +39,8 @@ else
     echo "${HOSTNAME} is configured with iSCSI"
 fi
 
+EXIT_VAL=0
+
 # Verifying whether Marshal agent is running or not and if there are any errors
 
 echo "*******************************"
@@ -48,8 +48,9 @@ echo "Marshal agent is $(systemctl is-active sbps-marshal.service)"
 
 ERR=$(journalctl -xeu sbps-marshal.service | tac | sed '/START SCAN/q' | grep -i error)
 
-if [ ! -z "${ERR}" ]; then
+if [ -n "${ERR}" ]; then
     echo "Marshal agent is having errors, please check"
+    EXIT_VAL=1
 fi
 echo "*******************************"
 
@@ -68,6 +69,7 @@ echo "*******************************"
 
 # Verify TCP service probes complete against all active portals (iSCSI)
 
+# shellcheck disable=SC2207
 PORTALS=($(ss -tnpl | grep ':3260 ' | awk '{ print $4 }' ))
 
 if [[ -n "${PORTALS[*]}" ]]; then
@@ -114,10 +116,9 @@ else
     exit 1
 fi
 
+# shellcheck disable=SC2207,SC2002
 SRV=($(cat tmp_file | awk '{print $(NF)}' | sed -e 's/\.$//'))
 
-A_REC=1
-
 echo "DNS A records are as below"
 echo "--------------------------"
 
@@ -126,7 +127,7 @@ do
     DNS_A=$(dig -t A +short "${i}")
     if [ -z "${DNS_A}" ]; then
         echo "ERROR: DNS A record is absent for ${i}"
-        A_REC=0
+        EXIT_VAL=1
     else
         echo "DNS A record is present: ${DNS_A} for ${i}"
     fi
@@ -144,8 +145,4 @@ for s in $(dig -t srv +short _sbps-hsn._tcp."${host}" | sort -k3 | awk '{print $
 
 for s in $(dig -t srv +short _sbps-nmn._tcp."${host}" | sort -k3 | awk '{print $NF;}' | xargs); do printf '** %s **\n' "$s"; dig +short "$s"; done
 
-if [ -z "$A_REC" ]; then
-    exit 1
-fi
-
-exit 0
+exit ${EXIT_VAL}
```

aasha-hpe:
> Here's a summary of the final changes I'm requesting. This includes changing the shellcheck disable from global to the file, to on the specific lines where they are needed.
> 
> ```
> --- a/goss-testing/scripts/iscsi_sbps_sanity.sh
> +++ b/goss-testing/scripts/iscsi_sbps_sanity.sh
> @@ -25,8 +25,6 @@
>  # This script is to do some sanity tests for iSCSI based boot content 
>  # projection.
>  
> -# shellcheck disable=SC2207,SC2143,SC2002
> -
>  LIO_ACT=0
>  
>  echo "Checking whether the node is configured with iSCSI"
> @@ -41,6 +39,8 @@ else
>      echo "${HOSTNAME} is configured with iSCSI"
>  fi
>  
> +EXIT_VAL=0
> +
>  # Verifying whether Marshal agent is running or not and if there are any errors
>  
>  echo "*******************************"
> @@ -48,8 +48,9 @@ echo "Marshal agent is $(systemctl is-active sbps-marshal.service)"
>  
>  ERR=$(journalctl -xeu sbps-marshal.service | tac | sed '/START SCAN/q' | grep -i error)
>  
> -if [ ! -z "${ERR}" ]; then
> +if [ -n "${ERR}" ]; then
>      echo "Marshal agent is having errors, please check"
> +    EXIT_VAL=1
>  fi
>  echo "*******************************"
>  
> @@ -68,6 +69,7 @@ echo "*******************************"
>  
>  # Verify TCP service probes complete against all active portals (iSCSI)
>  
> +# shellcheck disable=SC2207
>  PORTALS=($(ss -tnpl | grep ':3260 ' | awk '{ print $4 }' ))
>  
>  if [[ -n "${PORTALS[*]}" ]]; then
> @@ -114,10 +116,9 @@ else
>      exit 1
>  fi
>  
> +# shellcheck disable=SC2207,SC2002
>  SRV=($(cat tmp_file | awk '{print $(NF)}' | sed -e 's/\.$//'))
>  
> -A_REC=1
> -
>  echo "DNS A records are as below"
>  echo "--------------------------"
>  
> @@ -126,7 +127,7 @@ do
>      DNS_A=$(dig -t A +short "${i}")
>      if [ -z "${DNS_A}" ]; then
>          echo "ERROR: DNS A record is absent for ${i}"
> -        A_REC=0
> +        EXIT_VAL=1
>      else
>          echo "DNS A record is present: ${DNS_A} for ${i}"
>      fi
> @@ -144,8 +145,4 @@ for s in $(dig -t srv +short _sbps-hsn._tcp."${host}" | sort -k3 | awk '{print $
>  
>  for s in $(dig -t srv +short _sbps-nmn._tcp."${host}" | sort -k3 | awk '{print $NF;}' | xargs); do printf '** %s **\n' "$s"; dig +short "$s"; done
>  
> -if [ -z "$A_REC" ]; then
> -    exit 1
> -fi
> -
> -exit 0
> +exit ${EXIT_VAL}
> ```

Thanks a lot David !. I have incoporated the same and tested the same. Pls check. If no comments, then will squash the commits and push.

aasha-hpe:
> > Here's a summary of the final changes I'm requesting. This includes changing the shellcheck disable from global to the file, to on the specific lines where they are needed.
> > ```
> > --- a/goss-testing/scripts/iscsi_sbps_sanity.sh
> > +++ b/goss-testing/scripts/iscsi_sbps_sanity.sh
> > @@ -25,8 +25,6 @@
> >  # This script is to do some sanity tests for iSCSI based boot content 
> >  # projection.
> >  
> > -# shellcheck disable=SC2207,SC2143,SC2002
> > -
> >  LIO_ACT=0
> >  
> >  echo "Checking whether the node is configured with iSCSI"
> > @@ -41,6 +39,8 @@ else
> >      echo "${HOSTNAME} is configured with iSCSI"
> >  fi
> >  
> > +EXIT_VAL=0
> > +
> >  # Verifying whether Marshal agent is running or not and if there are any errors
> >  
> >  echo "*******************************"
> > @@ -48,8 +48,9 @@ echo "Marshal agent is $(systemctl is-active sbps-marshal.service)"
> >  
> >  ERR=$(journalctl -xeu sbps-marshal.service | tac | sed '/START SCAN/q' | grep -i error)
> >  
> > -if [ ! -z "${ERR}" ]; then
> > +if [ -n "${ERR}" ]; then
> >      echo "Marshal agent is having errors, please check"
> > +    EXIT_VAL=1
> >  fi
> >  echo "*******************************"
> >  
> > @@ -68,6 +69,7 @@ echo "*******************************"
> >  
> >  # Verify TCP service probes complete against all active portals (iSCSI)
> >  
> > +# shellcheck disable=SC2207
> >  PORTALS=($(ss -tnpl | grep ':3260 ' | awk '{ print $4 }' ))
> >  
> >  if [[ -n "${PORTALS[*]}" ]]; then
> > @@ -114,10 +116,9 @@ else
> >      exit 1
> >  fi
> >  
> > +# shellcheck disable=SC2207,SC2002
> >  SRV=($(cat tmp_file | awk '{print $(NF)}' | sed -e 's/\.$//'))
> >  
> > -A_REC=1
> > -
> >  echo "DNS A records are as below"
> >  echo "--------------------------"
> >  
> > @@ -126,7 +127,7 @@ do
> >      DNS_A=$(dig -t A +short "${i}")
> >      if [ -z "${DNS_A}" ]; then
> >          echo "ERROR: DNS A record is absent for ${i}"
> > -        A_REC=0
> > +        EXIT_VAL=1
> >      else
> >          echo "DNS A record is present: ${DNS_A} for ${i}"
> >      fi
> > @@ -144,8 +145,4 @@ for s in $(dig -t srv +short _sbps-hsn._tcp."${host}" | sort -k3 | awk '{print $
> >  
> >  for s in $(dig -t srv +short _sbps-nmn._tcp."${host}" | sort -k3 | awk '{print $NF;}' | xargs); do printf '** %s **\n' "$s"; dig +short "$s"; done
> >  
> > -if [ -z "$A_REC" ]; then
> > -    exit 1
> > -fi
> > -
> > -exit 0
> > +exit ${EXIT_VAL}
> > ```
> 
> Thanks a lot David !. I have incoporated the same and tested the same. Pls check. If no comments, then will squash the commits and push.

I just squashed the commits.

dborman-hpe (approved):
LGTM
Thanks for all your work on this!

don-bahls-hpe (approved):

aasha-hpe:
Thanks @don-bahls-hpe. @dborman-hpe hope this can be merged now.

PR #720, opened 2025-07-07
CRAYSAT-1955: Create functional tests for `sat hwinv`

annapoorna-s-alt:
## Summary and Scope

_Summarize what has changed. Explain why this PR is necessary. What is impacted? Is this a new feature, critical bug fix, etc?_

Create functional tests for `sat hwinv`  that run as part of the tests in csm-testing

## Issues and Related PRs

* Resolves [CRAYSAT-1955](https://jira-pro.it.hpe.com:8443/browse/CRAYSAT-1955)

## Testing

### Tested on:

  * rocket

### Test description:

- Tested with goss test downloading unstable rpm

## Risks and Mitigations

_Are there known issues with these changes? Any other special considerations?_
NO

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
+++ b/src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
@@ -1,0 +1,412 @@
+#
+# MIT License
+#
+# (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+"""
+Simple tests for the functionality of the `sat hwinv` command.
+"""
+import shlex
+import subprocess
+import unittest
+import re
+import json
+from typing import List
+
+
+HWINV_HEADERS = {
+    "nodes": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Cabinet Type', 'Memory Type', 'Memory Device Type', 'Memory Manufacturer',
+        'Memory Model', 'Memory Size (GiB)', 'Memory Module Count', 'Processor Count',
+        'Processor Manufacturer', 'Processor Model', 'Accelerator Count',
+        'Accelerator Riser Count', 'HSN NIC Count', 'Drive Count',
+        'Total Drive Capacity (GiB)', 'BIOS Version'
+    ],
+    "chassis": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
+    ],
+    "HSN boards": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
+    ],
+    "compute modules": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
+    ],
+    "router modules": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
+    ],
+    "node enclosures": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
+    ],
+    "node enclosure power supplies": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
+    ],
+    "processors": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Total Cores', 'Total Threads', 'Max Speed (MHz)'
+    ],
+    "node accelerators": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Location Name'
+    ],
+    "node accelerator risers": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Board Serial Number', 'Producer', 'Engineering Change Level'
+    ],
+    "node HSN NICS": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
+    ],
+    "memory modules": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Memory Type', 'Device Type', 'Capacity (MiB)', 'Operating Speed (MHz)'
+    ],
+    "drives": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Media Type', 'Capacity (GiB)', 'Percent Life Left'
+    ],
+    "CMM rectifiers": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Power Input Watts', 'Power Output Watts', 'Power Supply Type', 'Firmware Version'
+    ],
+    "node bmcs": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Manager Type', 'Firmware Version'
+    ],
+    "router bmcs": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Manager Type', 'Firmware Version'
+    ],
+    "mgmt switches": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Chassis Type'
+    ],
+    "cabinet pdus": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Equipment Type', 'Firmware Version'
+    ],
+    "cabinet pdu power connectors": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Nominal Voltage', 'Outlet Type', 'Phase Wiring Type', 'Power Enabled',
+        'Rated Current Amps', 'Voltage Type'
+    ]
+}
+
+
+SECTION_TO_TYPE = {
+    "nodes": "Node",
+    "chassis": "Chassis",
+    "HSN boards": "HSNBoard",
+    "compute modules": "ComputeModule",
+    "router modules": "RouterModule",
+    "node enclosures": "NodeEnclosure",
+    "node enclosure power supplies": "NodeEnclosurePowerSupply",
+    "processors": "Processor",
+    "node accelerators": "NodeAccel",
+    "node accelerator risers": "NodeAccelRiser",
+    "node HSN NICS": "NodeHsnNic",
+    "memory modules": "Memory",
+    "drives": "Drive",
+    "CMM rectifiers": "CMMRectifier",
+    "node bmcs": "NodeBMC",
+    "router bmcs": "RouterBMC",
+    "mgmt switches": "MgmtSwitch",
+    "cabinet pdus": "CabinetPDU",
+    "cabinet pdu power connectors": "CabinetPDUPowerConnector"
+}
+
+
+def get_present_components() -> List[str]:
+    """Retrieve the list of components present in the system using HSM.
+    
+    Returns:
+        The list of unique component types present in the system.
+    """
+    command = "cray hsm inventory hardware list --format json"
+    proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    hsm_output = proc.stdout.decode()
+    components = json.loads(hsm_output)
+    return list(set([component["Type"] for component in components]))
+
+
+def adjust_expected_header(header: List[str], sat_output: str) -> List[str]:
+    """Return a copy of the specified header, removing keys which were omitted in the output"""
+    new_header = header.copy()
+    for col in header:
+        info_miss = f"INFO: All values for '{col}' are 'MISSING', omitting key."
+        info_empty = f"INFO: All values for '{col}' are 'EMPTY', omitting key."
+        if any(msg in sat_output for msg in (info_miss, info_empty)):
+            new_header.remove(col)
+    return new_header
+
+
+def get_column_names_list(input_string: str) -> List[str]:
+    """Return the list of column names specified in the header line of the input string."""
+    lines = input_string.strip().split('\n')
+    for line in lines:
+        if line.strip().startswith('| xname'):
+            header_line = line
+            return [col.strip() for col in header_line.split('|') if col.strip()]
+    return []
+
+
+class TestHwinv(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.present_components = get_present_components()
+
+    def run_hwinv_test(self, command: str, expected_key: str, adjust_headers: bool = True) -> None:
+        """Helper function to run an hwinv command test."""
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        stderr = proc.stderr.decode()
+        self.assertIn(f"Listing of all {expected_key}", stdout)
+        actual_header = get_column_names_list(stdout)
+        if adjust_headers:
+            expected_header = adjust_expected_header(HWINV_HEADERS[expected_key], stderr)
+        else:
+            expected_header = HWINV_HEADERS[expected_key]
+
+        hsm_type = SECTION_TO_TYPE[expected_key]
+        if not actual_header:
+            if hsm_type in self.present_components:
+                self.fail(f'The table for {expected_key} is empty, but there are components of '
+                          f'{hsm_type} type present in the system.')
+            else:
+                # If there are not components of this type, the table is expected to be empty
+                return
+        self.assertEqual(expected_header, actual_header)
+
+    def test_hwinv_list_nodes(self) -> None:
+        """Test that `sat hwinv --list-nodes` returns the proper header."""
+        command = 'sat hwinv --list-nodes'
+        self.run_hwinv_test(command, 'nodes')
+
+    def test_hwinv_list_nodes_include_miss_and_empty(self) -> None:
+        """Test that `sat hwinv --list-nodes --show-missing --show-empty` returns all the header."""
+        command = 'sat hwinv --list-nodes --show-missing --show-empty'
+        self.run_hwinv_test(command, 'nodes', adjust_headers=False)
+
+    def test_hwinv_list_drives_include_miss_and_empty(self) -> None:
+        """Test that `sat hwinv --list-drives --show-missing --show-empty` returns all the header."""
+        command = 'sat hwinv --list-drives --show-missing --show-empty'
+        self.run_hwinv_test(command, 'drives', adjust_headers=False)
+
+    def test_hwinv_list_chassis(self) -> None:
+        """Test that `sat hwinv --list-chassis` returns proper header."""
+        command = 'sat hwinv --list-chassis'
+        self.run_hwinv_test(command, 'chassis')
+
+    def test_hwinv_list_hsnboards(self) -> None:
+        """Test that `sat hwinv --list-hsn-boards` returns proper header"""
+        command = 'sat hwinv --list-hsn-boards'
+        self.run_hwinv_test(command, 'HSN boards')
+
+    def test_hwinv_list_compute_modules(self) -> None:
+        """Test that `sat hwinv --list-compute-modules` returns proper header"""
+        command = 'sat hwinv --list-compute-modules'
+        self.run_hwinv_test(command, 'compute modules')
+
+    def test_hwinv_list_router_modules(self) -> None:
+        """Test that `sat hwinv --list-router-modules` returns proper header"""
+        command = 'sat hwinv --list-router-modules'
+        self.run_hwinv_test(command, 'router modules')
+
+    def test_hwinv_list_node_enclosures(self) -> None:
+        """Test that `sat hwinv --list-node-enclosures` returns proper header"""
+        command = 'sat hwinv --list-node-enclosures'
+        self.run_hwinv_test(command, 'node enclosures')
+
+    def test_hwinv_list_node_enclosure_power_supplies(self) -> None:
+        """Test that `sat hwinv --list-node-enclosure-power-supplies` returns proper header"""
+        command = 'sat hwinv --list-node-enclosure-power-supplies'
+        self.run_hwinv_test(command, 'node enclosure power supplies')
+
+    def test_hwinv_list_procs(self) -> None:
+        """Test that `sat hwinv --list-procs` returns proper header"""
+        command = 'sat hwinv --list-procs'
+        self.run_hwinv_test(command, 'processors')
+
+    def test_hwinv_list_node_accels(self) -> None:
+        """Test that `sat hwinv --list-node-accels` returns proper header"""
+        command = 'sat hwinv --list-node-accels'
+        self.run_hwinv_test(command, 'node accelerators')
+
+    def test_hwinv_list_node_accel_risers(self) -> None:
+        """Test that `sat hwinv --list-node-accel-risers` returns proper header"""
+        command = 'sat hwinv --list-node-accel-risers'
+        self.run_hwinv_test(command, 'node accelerator risers')
+
+    def test_hwinv_list_node_hsn_nics(self) -> None:
+        """Test that `sat hwinv --list-node-hsn-nics` returns proper header"""
+        command = 'sat hwinv --list-node-hsn-nics'
+        self.run_hwinv_test(command, 'node HSN NICS')
+
+    def test_hwinv_list_mems(self) -> None:
+        """Test that `sat hwinv --list-mems` returns proper header"""
+        command = 'sat hwinv --list-mems'
+        self.run_hwinv_test(command, 'memory modules')
+
+    def test_hwinv_list_drives(self) -> None:
+        """Test that `sat hwinv --list-drives` returns proper header"""
+        command = 'sat hwinv --list-drives'
+        self.run_hwinv_test(command, 'drives')
+
+    def test_hwinv_list_cmm_rectifiers(self) -> None:
+        """Test that `sat hwinv --list-cmm-rectifiers` returns proper header"""
+        command = 'sat hwinv --list-cmm-rectifiers'
+        self.run_hwinv_test(command, 'CMM rectifiers')
+
+    def test_hwinv_list_node_bmcs(self) -> None:
+        """Test that `sat hwinv --list-node-bmcs` returns proper header"""
+        command = 'sat hwinv --list-node-bmcs'
+        self.run_hwinv_test(command, 'node bmcs')
+
+    def test_hwinv_list_router_bmcs(self) -> None:
+        """Test that `sat hwinv --list-router-bmcs` returns proper header"""
+        command = 'sat hwinv --list-router-bmcs'
+        self.run_hwinv_test(command, 'router bmcs')
+
+    def test_hwinv_list_mgmt_switches(self) -> None:
+        """Test that `sat hwinv --list-mgmt-switches` returns proper header"""
+        command = 'sat hwinv --list-mgmt-switches'
+        self.run_hwinv_test(command, 'mgmt switches')
+
+    def test_hwinv_list_cabinet_pdus(self) -> None:
+        """Test that `sat hwinv --list-cabinet-pdus` returns proper header"""
+        command = 'sat hwinv --list-cabinet-pdus'
+        self.run_hwinv_test(command, 'cabinet pdus')
+
+    def test_hwinv_list_cabinet_pdu_power_connectors(self) -> None:
+        """Test that `sat hwinv --list-cabinet-pdu-power-connectors` returns proper header"""
+        command = 'sat hwinv --list-cabinet-pdu-power-connectors'
+        self.run_hwinv_test(command, 'cabinet pdu power connectors')
+
+    def test_hwinv_list_all(self) -> None:
+        """Test that `sat hwinv --list-all --show-missing --show-empty` validates only the components present in the system."""
+        command = 'sat hwinv --list-all --show-missing --show-empty'
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        stdout_lines = stdout.splitlines()
+
+        # Iterate through the lines to find sections and validate headers
+        for i, line in enumerate(stdout_lines):
+            # Match section titles like "Listing of all <SOME COMPONENT TYPE>"
+            match = re.match(r"Listing of all (.+)", line.strip())
+            if match:
+                section_name = match.group(1)
+                component_type = SECTION_TO_TYPE.get(section_name)
+
+                # If the section doesn't correspond to any present component, skip it
+                if component_type not in self.present_components:
+                    continue
+
+                self.assertIn(section_name, HWINV_HEADERS,
+                    f"Unexpected section '{section_name}' found in output."
+                )
+
+                expected_headers = HWINV_HEADERS[section_name]
+
+                # Get the header line (3 lines below the section title)
+                header_line_index = i + 3
+                if header_line_index >= len(stdout_lines):
+                    self.fail(f"Header row for section '{section_name}' not found in output.")
+                header_line = stdout_lines[header_line_index].strip()
+
+                # Parse the actual header
+                actual_header = [col.strip() for col in header_line.split("|") if col.strip()]
+
+                # Validate the actual header against the adjusted expected header
+                self.assertEqual(expected_headers, actual_header,
+                    f"Header for section '{section_name}' does not match the expected fields. "
+                    f"Expected: {expected_headers}, Actual: {actual_header}"
+                )
+
+    def test_hwinv_list_node_fields(self) -> None:
+        """Test that `sat hwinv --list-nodes --node-fields 'xname, FRUID, Serial Number, SKU'` returns the proper header."""
+        expected_header = ['xname', 'FRUID', 'Serial Number', 'SKU']
+        command = f"sat hwinv --list-nodes --node-fields '{','.join(expected_header)}'"
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+
+        self.assertIn("Listing of all nodes", stdout)
+        actual_header = get_column_names_list(stdout)
+        if not actual_header:
+            return
+
+        self.assertEqual(expected_header, actual_header)
+
+    def test_hwinv_list_proc_fields(self) -> None:
+        """Test that `sat hwinv --list-procs --proc-field 'xname, FRUID, Manufacturer, Model'` returns proper header"""
+        expected_header = ['xname', 'FRUID', 'Manufacturer', 'Model']
+        command = f"sat hwinv --list-procs --proc-fields '{','.join(expected_header)}'"
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+
+        self.assertIn("Listing of all processors", stdout)
+        actual_header = get_column_names_list(stdout)
+        if not actual_header:
+            return
+
+        self.assertEqual(expected_header, actual_header)
+
+    def test_hwinv_field_limiting(self) -> None:
+        """Test that `sat hwinv --fields 'xname, FRUID, Part Number, SKU'` limits fields for all printed reports."""
+        expected_header = ['xname', 'FRUID', 'Part Number', 'SKU']
+        command = f"sat hwinv --fields '{','.join(expected_header)}'"
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        stdout_lines = stdout.splitlines()
+
+        # Iterate through the lines to find sections and validate headers
+        for i, line in enumerate(stdout_lines):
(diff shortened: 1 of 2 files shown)


annapoorna-s-alt:
```
ncn-m001:/mnt/developer/sann/automation-test/hwinv # ./opt/cray/tests/install/csm-testing-python-venv/bin/test_sat_functional hwinv.test_hwinv
test_hwinv_field_limiting (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --fields` limits fields for all printed reports. ... ok
test_hwinv_list_all (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-all` prints all expected sections and their headers. ... ok
test_hwinv_list_cabinet_pdu_power_connectors (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-cabinet-pdu-power-connectors` returns proper header ... ok
test_hwinv_list_cabinet_pdus (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-cabinet-pdus` returns proper header ... ok
test_hwinv_list_chassis (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-chassis` returns proper header. ... ok
test_hwinv_list_cmm_rectifiers (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-cmm-rectifiers` returns proper header ... ok
test_hwinv_list_compute_modules (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-compute-modules` returns proper header ... ok
test_hwinv_list_drives (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-drives` returns proper header ... ok
test_hwinv_list_hsnboards (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-hsn-boards` returns proper header ... ok
test_hwinv_list_mems (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-mems` returns proper header ... ok
test_hwinv_list_mgmt_switches (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-mgmt-switches` returns proper header ... ok
test_hwinv_list_node_accel_risers (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-accel-risers` returns proper header ... ok
test_hwinv_list_node_accels (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-accels` returns proper header ... ok
test_hwinv_list_node_bmcs (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-bmcs` returns proper header ... ok
test_hwinv_list_node_enclosure_power_supplies (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-enclosure-power-supplies` returns proper header ... ok
test_hwinv_list_node_enclosures (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-enclosures` returns proper header ... ok
test_hwinv_list_node_fields (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-nodes --node-fields 'xname, FRUID, SKU, Serial Number'` returns the proper header. ... ok
test_hwinv_list_node_hsn_nics (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-hsn-nics` returns proper header ... ok
test_hwinv_list_nodes (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-nodes` returns the proper header. ... ok
test_hwinv_list_proc_fields (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-procs --proc-field 'xname, FRUID, Manufacturer, Model'` returns proper header ... ok
test_hwinv_list_procs (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-procs` returns proper header ... ok
test_hwinv_list_router_bmcs (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-router-bmcs` returns proper header ... ok
test_hwinv_list_router_modules (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-router-modules` returns proper header ... ok

----------------------------------------------------------------------
Ran 23 tests in 58.786s

OK

```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'Serial Number', 'SKU',
+    'Equipment Type', 'Firmware Version'
+]
+HWINV_CABINET_PDU_POWER_CONNECTORS_HEADER = [
+    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'Serial Number', 'SKU',
+    'Nominal Voltage', 'Outlet Type', 'Phase Wiring Type', 'Power Enabled',
+    'Rated Current Amps', 'Voltage Type'
+]
+
+def adjust_expected_header(header: List[str], sat_output: str) -> List[str]:
```
haasken-hpe, line 149: PEP8 wants an additional blank line

```suggestion

def adjust_expected_header(header: List[str], sat_output: str) -> List[str]:
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+    """Return a copy of the specified header, removing keys which were omitted in the output"""
+    new_header = header.copy()
+    for col in header:
+        info_miss = f"INFO: All values for '{col}' are 'MISSING', omitting key."
+        info_empty = f"INFO: All values for '{col}' are 'EMPTY', omitting key."
+        if any(msg in sat_output for msg in (info_miss, info_empty)):
+            new_header.remove(col)
+    return new_header
+
+def get_column_names_list(input_string: str) -> List[str]:
```
haasken-hpe, line 160:
```suggestion

def get_column_names_list(input_string: str) -> List[str]:
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+    """Return the list of column names specified in the header line of the input string."""
+    lines = input_string.strip().split('\n')
+    for line in lines:
+        if line.strip().startswith('| xname'):
+            header_line = line
+            return [col.strip() for col in header_line.split('|') if col.strip()]
+    return []
+
+class TestHwinv(unittest.TestCase):
```
haasken-hpe, line 170:
```suggestion

class TestHwinv(unittest.TestCase):
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+        """Test that `sat hwinv --list-nodes` returns the proper header."""
+        command = 'sat hwinv --list-nodes'
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        stderr = proc.stderr.decode()
+        all_output = stdout + '\n' + stderr
+
+        self.assertIn("Listing of all nodes", stdout)
+        actual_header = get_column_names_list(stdout)
+        expected_header = adjust_expected_header(HWINV_NODE_HEADER, all_output)
```
haasken-hpe: I think the info messages you're looking for should always go to stderr, so you could just check `stderr`. THen I think there's no need for `all_output` either.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+import unittest
+from typing import List
+
+HWINV_NODE_HEADER = [
+    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+    'Cabinet Type', 'Memory Type', 'Memory Device Type', 'Memory Manufacturer',
+    'Memory Size (GiB)', 'Memory Module Count', 'Processor Count', 'Processor Manufacturer',
+    'Processor Model', 'Accelerator Count', 'Accelerator Riser Count', 'HSN NIC Count',
+    'Drive Count', 'Total Drive Capacity (GiB)', 'BIOS Version', 'Memory Model'
+]
```
haasken-hpe: Order does appear to matter to your tests, so make sure the order of these fields is correct. I noticed that "Memory Model" is in the wrong position in this one.

```
HWINV_NODE_HEADER = [
    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
    'Cabinet Type', 'Memory Type', 'Memory Device Type', 'Memory Manufacturer', 'Memory Model'
    'Memory Size (GiB)', 'Memory Module Count', 'Processor Count', 'Processor Manufacturer',
    'Processor Model', 'Accelerator Count', 'Accelerator Riser Count', 'HSN NIC Count',
    'Drive Count', 'Total Drive Capacity (GiB)', 'BIOS Version'
]
```

A quick way to verify this is to use `--show-empty` and `--show-missing` and look at the colums.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+HWINV_CHASSIS_HEADER = [
+    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'Serial Number', 'SKU'
```
haasken-hpe: It looks like SKU always comes before "Serial Number":

```suggestion
    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number'
```

haasken-hpe: I won't comment on the other instances of this, but it appears to be the same throughout.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'Serial Number', 'SKU',
+    'Memory Type', 'Device Type', 'Capacity (MiB)', 'Operating Speed (MHz)'
+]
+HWINV_DRIVES_HEADER = [
+    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'Serial Number', 'SKU',
+    'Media Type', 'Capacity (GiB)', 'Percent Life Left'
+]
+HWINV_CMM_RECTIFIERS_HEADER = [
+    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'Serial Number', 'SKU',
+    'Power Supply Type', 'Firmware Version', 'Power Input Watts', 'Power Output Watts'
```
haasken-hpe: I think this is the correct order:

```suggestion
    'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
    'Power Input Watts', 'Power Output Watts', 'Power Supply Type', 'Firmware Version'
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+                        for j in range(i+1, len(lines)):
+                            next_line = lines[j].strip()
+                            if next_line.startswith('|') and 'xname' in next_line:
+                                header_line = lines[j]
+                                found_table = True
+                                break
+                        break
+                if not found_table or header_line is None:
+                    # No table for this section, skip header check
+                    continue
```
haasken-hpe: I think this occurs if there are no components of this type. Is that correct? If so, we might want to check that there are indeed no items of this type to ensure that the table is not missing for some other reason like a bug in `sat`. I think we'd have to check the count of the number of components in HSM by querying the HSM API through `cray hsm` or something. I'm not sure if it's worth it, but something to consider.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+            "Listing of all node bmcs": HWINV_NODE_BMCS_HEADER,
+            "Listing of all router bmcs": HWINV_ROUTER_BMCS_HEADER,
+            "Listing of all mgmt switches": HWINV_MGMT_SWITCHES_HEADER,
+            "Listing of all cabinet pdus": HWINV_CABINET_PDUS_HEADER,
+            "Listing of all cabinet pdu power connectors": HWINV_CABINET_PDU_POWER_CONNECTORS_HEADER
+        }
+        for section, expected_header in section_headers.items():
+            with self.subTest(section=section):
+                self.assertIn(section, stdout, f"Section title '{section}' not found in output.")
+                section_start = stdout.find(section)
```
haasken-hpe: You could probably simplify this code a bit if you split `stdout` into lines and iterated over the lines of output instead of just processing it as one big multi-line string.

If you split the stdout into lines, you could just walk through the lines, and each time you find the heading of the form:

```
################################################################################
Listing of all <SOME COMPONENT TYPE>
################################################################################
```

You could then look up the fields for that hardware type. E.g., you could do a regex match and extract the group. Then you would further be able to know that the header row of the table should appear 3 lines further down in the output.

This approach would also have lower time complexity because you would only iterate over the lines of the output once (I guess twice if you do `splitlines` first) as opposed to searching linearly (via the `find` call) through the output for every hardware type we are looking for.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+                if section_start == -1:
+                    self.fail(f"Section '{section}' not found in output.")
+                lines = stdout[section_start:]
+                header_line = None
+                found_table = False
+                for i, line in enumerate(lines):
```
haasken-hpe: You appear to be treating `line` like it's a single line of output, but it looks like it's actually the entire `stdout` string from `section_start` and on. Am I missing something?

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+
+    def test_hwinv_field_limiting(self) -> None:
+        """Test that `sat hwinv --fields` limits fields for all printed reports."""
+        selected_fields = 'xname, FRUID, Part Number, SKU'
+        command = f"sat hwinv --fields '{selected_fields}'"
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+
+        # Fields we expect in the output based on the `--fields` option
+        expected_header = ['xname', 'FRUID', 'Part Number', 'SKU']
```
haasken-hpe: Could use `selected_fields` to avoid repeating it.

haasken-hpe (changes requested):
The coverage of the test cases from the Jira looks good.

An additional test that would be easy to add is to run at least one of these `--list` options with `--show-missing`, `--show-empty`, or perhaps both.

I am not confident that the logic in `test_hwinv_list_all` is correct. Can you take a  look at my comments and perhaps try to simplify it? There's also a lot of duplicate complicated logic between `test_hwinv_list_all` and `test_hwinv_field_limiting`.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,630 @@
...
+
+                # Validate that fields not in `selected_fields` are excluded
+                excluded_fields = set(full_header) - set(expected_header)
+                for excluded_field in excluded_fields:
+                    self.assertNotIn(
+                        excluded_field, actual_header,
+                        f"Excluded field '{excluded_field}' found in header for section '{section}'."
+                    )
+
+if __name__ == "__main__":
```
haasken-hpe, line 410: PEP8

```suggestion

if __name__ == "__main__":
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
annapoorna-s-alt: Yes, my bad. i should have taken order using `--show-empty` and `--show-missing`. Addressed

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
annapoorna-s-alt: Thanks for noticing. I agree that all info messages goes to stderr and we can get rid of `all_output`

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
annapoorna-s-alt: Will remove selected fields and keep expected_headers since I'm not using selected fields anywhere other than in a command.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,632 @@
...
+
+        # Split the output into lines
+        lines = stdout.splitlines()
+        line_count = len(lines)
+
+        # Iterate through lines and process each section
+        for i in range(line_count):
+            line = lines[i].strip()
+
+            if line.startswith("Listing of all ") and line.endswith("################################################################################"):
```
haasken-hpe: This condition is never true because "Listing of all" is on one line, and the "######" string is on the next line, not the same line.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,632 @@
...
+
+    def test_hwinv_list_all(self) -> None:
+        """Test that `sat hwinv --list-all` prints all expected sections and their headers."""
+        command = 'sat hwinv --list-all'
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        stderr = proc.stderr.decode()
+
+        section_headers = {
+            "Listing of all nodes": HWINV_NODE_HEADER,
```
haasken-hpe: I would suggest removing the repeated "Listing of all" part of the keys and just use the human-friendly component type. E.g.:

```suggestion
            "nodes": HWINV_NODE_HEADER,
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,632 @@
...
+            if line.startswith("Listing of all ") and line.endswith("################################################################################"):
+                section_title = re.match(r"Listing of all (.+)", line).group(1)
+                self.assertIn(
+                    section_title,
+                    section_headers,
+                    f"Unexpected section title '{section_title}' found in output."
+                )
+
+                # Extract the expected header for this section
+                expected_header = section_headers[section_title]
```
haasken-hpe: Right now, this will give a KeyError because the keys include "Listing of all", but I suggested changing that above.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,632 @@
...
+                )
+
+                # Validate that excluded fields are not present in the actual header
+                excluded_fields = set(expected_header) - set(adjusted_header)
+                for excluded_field in excluded_fields:
+                    self.assertNotIn(
+                        excluded_field,
+                        actual_header,
+                        f"Excluded field '{excluded_field}' found in header for section '{section_title}'."
+                    )
```
haasken-hpe: This is redundant with the equality check on line 514.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,632 @@
...
+        # Split the output into lines
+        lines = stdout.splitlines()
+        line_count = len(lines)
+
+        # Iterate through lines and process each section
+        for i in range(line_count):
+            line = lines[i].strip()
+
+            if line.startswith("Listing of all ") and line.endswith("################################################################################"):
+                section_title = re.match(r"Listing of all (.+)", line).group(1)
```
haasken-hpe: We're not yet checking that every expected section appears in the output. You could do that by saving these `section_title` as they're found. Then at the end of the loop, check that every key in `section_headers` has been found.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,632 @@
...
+                # Extract the expected header for this section
+                expected_header = section_headers[section_title]
+
+                # Skip to the header row (3 lines down)
+                header_line_index = i + 3
+                if header_line_index >= line_count:
+                    self.fail(f"Header row for section '{section_title}' not found in output.")
+                header_line = lines[header_line_index].strip()
+
+                # Ensure the header line starts with "|" and contains "xname"
```
haasken-hpe: Also need to account for the possibility of empty tables for components that don't exist on the system like you did in your other tests.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,632 @@
...
+        actual_header = get_column_names_list(stdout)
+        expected_header = adjust_expected_header(HWINV_NODE_HEADER, stderr)
+
+        if not actual_header:
+            return
+
+        self.assertEqual(expected_header, actual_header)
+
+    def test_hwinv_list_include_miss_and_empty(self) -> None:
+        """Test that `sat hwinv --list-nodes --show-missing --show-empty` returns all the header."""
```
haasken-hpe, line 204: This is a good test case, but I would suggest adding another just like it but for another type that usually has a missing or empty field (e.g. `--list-hsn-boards`).

haasken-hpe (changes requested):

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
annapoorna-s-alt: drives has more empty fields. Other than that I added in `--list-all` as well where we get bunch of missing and empty

annapoorna-s-alt:
Testing on rocket:

```
ncn-m001:/mnt/developer/sann/automation-test/hwinv/CRAYSAT-1955/28-7 # ./opt/cray/tests/install/csm-testing-python-venv/bin/test_sat_functional hwinv.test_hwinv
test_hwinv_field_limiting (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --fields 'xname, FRUID, Part Number, SKU'` limits fields for all printed reports. ... ok
test_hwinv_list_all (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-all --show-missing --show-empty` validates only the components present in the system. ... ok
test_hwinv_list_cabinet_pdu_power_connectors (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-cabinet-pdu-power-connectors` returns proper header ... ok
test_hwinv_list_cabinet_pdus (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-cabinet-pdus` returns proper header ... ok
test_hwinv_list_chassis (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-chassis` returns proper header. ... ok
test_hwinv_list_cmm_rectifiers (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-cmm-rectifiers` returns proper header ... ok
test_hwinv_list_compute_modules (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-compute-modules` returns proper header ... ok
test_hwinv_list_drives (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-drives` returns proper header ... ok
test_hwinv_list_drives_include_miss_and_empty (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-drives --show-missing --show-empty` returns all the header. ... ok
test_hwinv_list_hsnboards (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-hsn-boards` returns proper header ... ok
test_hwinv_list_mems (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-mems` returns proper header ... ok
test_hwinv_list_mgmt_switches (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-mgmt-switches` returns proper header ... ok
test_hwinv_list_node_accel_risers (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-accel-risers` returns proper header ... ok
test_hwinv_list_node_accels (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-accels` returns proper header ... ok
test_hwinv_list_node_bmcs (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-bmcs` returns proper header ... ok
test_hwinv_list_node_enclosure_power_supplies (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-enclosure-power-supplies` returns proper header ... ok
test_hwinv_list_node_enclosures (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-enclosures` returns proper header ... ok
test_hwinv_list_node_fields (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-nodes --node-fields 'xname, FRUID, Serial Number, SKU'` returns the proper header. ... ok
test_hwinv_list_node_hsn_nics (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-node-hsn-nics` returns proper header ... ok
test_hwinv_list_nodes (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-nodes` returns the proper header. ... ok
test_hwinv_list_nodes_include_miss_and_empty (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-nodes --show-missing --show-empty` returns all the header. ... ok
test_hwinv_list_proc_fields (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-procs --proc-field 'xname, FRUID, Manufacturer, Model'` returns proper header ... ok
test_hwinv_list_procs (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-procs` returns proper header ... ok
test_hwinv_list_router_bmcs (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-router-bmcs` returns proper header ... ok
test_hwinv_list_router_modules (csm_testing.tests.sat_functional.hwinv.test_hwinv.TestHwinv)
Test that `sat hwinv --list-router-modules` returns proper header ... ok

----------------------------------------------------------------------
Ran 25 tests in 69.039s

OK

```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+    "cabinet pdu power connectors": [
+        'xname', 'FRUID', 'Manufacturer', 'Model', 'Part Number', 'SKU', 'Serial Number',
+        'Nominal Voltage', 'Outlet Type', 'Phase Wiring Type', 'Power Enabled',
+        'Rated Current Amps', 'Voltage Type'
+    ]
+}
+
+
+def get_present_components() -> List[str]:
+    """Retrieve the list of components present in the system using HSM."""
```
haasken-hpe:
```suggestion
def get_present_component_types() -> List[str]:
    """Retrieve the list of component types present in the system using HSM.

    Returns:
        The list of unique component types present in the system.
    """
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def test_hwinv_list_nodes(self) -> None:
+        """Test that `sat hwinv --list-nodes` returns the proper header."""
+        command = 'sat hwinv --list-nodes'
+        self.run_hwinv_test(command, 'nodes')
+
+    def test_hwinv_list_nodes_include_miss_and_empty(self) -> None:
+        """Test that `sat hwinv --list-nodes --show-missing --show-empty` returns all the header."""
+        command = 'sat hwinv --list-nodes --show-missing --show-empty'
+        self.run_hwinv_test(command, 'nodes')
```
haasken-hpe: This test should not allow the adjustment of the expected headers. All the fields should appear when we specify `--show-missing` and `--show-empty`.

You could pretty easily add an argument to `run_hwinv_test`, e.g. `adjust_headers`, have it default to `True`, and then if it's True, adjust the headers, otherwise leave expected headers as the full list from `HWINV_HEADERS`. Then pass `False` for this new argument in this test and the next one.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+
+        # Iterate through the lines to find sections and validate headers
+        for i, line in enumerate(stdout_lines):
+            # Match section titles like "Listing of all <SOME COMPONENT TYPE>"
+            match = re.match(r"Listing of all (.+)", line.strip())
+            if not match:
+                continue
+
+            section_name = match.group(1)
+            lower_sec   = section_name.lower()
```
haasken-hpe:
```suggestion
            lower_sec = section_name.lower()
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+            # Skip any section type not reported as present
+            if lower_sec not in present_components:
```
haasken-hpe: The section names are human-friendly names from the `sat hwinv` code, which are lower case except for the initialisms (e.g. BMC), and the words are separated by spaces. They do not exactly match the types in HSM, which are Upper CamelCase and often abbreviated. Thus, I believe this if is always true, and it will always continue without executing the rest of the loop.

I think the easiest way to address this is to provide a hard-coded mapping from the `sat hwinv` section titles to the `HSM` types.

To illustrate more clearly, here are the types from HSM:

```
# cray hsm inventory hardware list | jq -r '.[].Type' | sort -u
Chassis
CMMRectifier
ComputeModule
Drive
HSNBoard
Memory
MgmtSwitch
Node
NodeAccel
NodeAccelRiser
NodeBMC
NodeEnclosure
NodeEnclosurePowerSupply
NodeHsnNic
Processor
RouterBMC
RouterModule
```

Compared to the headings from `sat hwinv`:

```
# sat hwinv --list-all 2>/dev/null | grep "Listing of all" | sed 's/Listing of all \(\S\+\)/\1/'
chassis
CMM rectifiers
compute modules
drives
HSN boards
memory modules
nodes
node enclosures
node enclosure power supplies
processors
node accelerators
node accelerator risers
node HSN NICS
router modules
node bmcs
router bmcs
mgmt switches
cabinet pdus
cabinet pdu power connectors
```

Note that it's not even sufficient to make both lowercase and remove spaces because of abbreviated types in HSM like "NodeAccel" and "NodeAccelRiser" which are spelled out and made plural in the `sat hwinv` listing section headers.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+                f"Header for section '{section_name}' does not match the expected fields. "
+                f"Expected: {expected_header}, Actual: {actual_header}"
+            )
+
+    def test_hwinv_list_node_fields(self) -> None:
+        """Test that `sat hwinv --list-nodes --node-fields 'xname, FRUID, Serial Number, SKU'` returns the proper header."""
+        command = "sat hwinv --list-nodes --node-fields 'xname, FRUID, Serial Number, SKU'"
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        expected_header = ['xname', 'FRUID', 'Serial Number', 'SKU']
```
haasken-hpe: Minor suggestion to make it unnecessary to specify the fields twice:

```suggestion
        expected_header = ['xname', 'FRUID', 'Serial Number', 'SKU']
        command = f"sat hwinv --list-nodes --node-fields '{",".join(expected_header)}'"
        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
        stdout = proc.stdout.decode()
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+    def run_hwinv_test(self, command: str, expected_key: str) -> None:
+        """Helper function to run an hwinv command test."""
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        stderr = proc.stderr.decode()
+        self.assertIn(f"Listing of all {expected_key}", stdout)
+        actual_header = get_column_names_list(stdout)
+        expected_header = adjust_expected_header(HWINV_HEADERS[expected_key], stderr)
+        if not actual_header:
+            return
```
haasken-hpe: If there are components of the type, we should expect there to be a table with content. We could store the list of component types from HSM (e.g. in the `setUpClass` method), and then we can use the mapping that from `sat hwinv` section headers to HSM component types to figure out if there should be a table with content.

Note that the logic you have here is essentially also the same as the logic you want in the two `--list-all` test cases, but in those test cases, you are performing that logic multiple times, i.e., once for each section of the output.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+
+            # Get the header line (3 lines below the section title)
+            header_line_index = i + 3
+            if header_line_index >= len(stdout_lines):
+                self.fail(f"Header row for section '{section_name}' not found in output.")
+            header_line = stdout_lines[header_line_index].strip()
+
+            # Check for empty tables (no header or data rows)
+            if not header_line.startswith("|"):
+                continue
```
haasken-hpe: If there are any components of the given type in the output of `cray hsm inventory hardware list`, then we should expect there to be a non-empty table here, and we should fail if the table is missing.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+            return
+
+        self.assertEqual(expected_header, actual_header)
+
+    def test_hwinv_list_proc_fields(self) -> None:
+        """Test that `sat hwinv --list-procs --proc-field 'xname, FRUID, Manufacturer, Model'` returns proper header"""
+        command = "sat hwinv --list-procs --proc-field 'xname, FRUID, Manufacturer, Model'"
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        expected_header = ['xname', 'FRUID', 'Manufacturer', 'Model']
```
haasken-hpe: Same idea here:

```suggestion
        expected_header = ['xname', 'FRUID', 'Manufacturer', 'Model']
        command = f"sat hwinv --list-procs --proc-fields '{",".join(expected_header)}'"
        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
        stdout = proc.stdout.decode()
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+    def test_hwinv_field_limiting(self) -> None:
+        """Test that `sat hwinv --fields 'xname, FRUID, Part Number, SKU'` limits fields for all printed reports."""
+        present_components = get_present_components()
+        command = f"sat hwinv --fields 'xname, FRUID, Part Number, SKU'"
+        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+        stdout = proc.stdout.decode()
+        stdout_lines = stdout.splitlines()
+
+        # Fields we expect in the output based on the `--fields` option
+        expected_header = ['xname', 'FRUID', 'Part Number', 'SKU']
```
haasken-hpe: Same idea here:

```suggestion
        expected_header = ['xname', 'FRUID', 'Part Number', 'SKU']
        command = f"sat hwinv --fields '{",".join(expected_header)}'"
        proc = subprocess.run(shlex.split(command), stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
        stdout = proc.stdout.decode()
        stdout_lines = stdout.splitlines()
```

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+
+            section_name = match.group(1)
+            lower_sec   = section_name.lower()
+            # Skip any section type not reported as present
+            if lower_sec not in present_components:
+                continue
+
+            self.assertIn(section_name, HWINV_HEADERS,
+                f"Unexpected section '{section_name}' found in output."
+            )
```
haasken-hpe: This code is exactly the same as lines 274-287, which suggests it could be refactored.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,389 @@
...
+                continue
+
+            # Parse the actual header
+            actual_header = [col.strip() for col in header_line.split("|") if col.strip()]
+
+            # Validate the actual header against the adjusted expected header
+            self.assertEqual( expected_header, actual_header,
+                f"Header for section '{section_name}' does not match the expected fields. "
+                f"Expected: {expected_header}, Actual: {actual_header}"
+            )
```
haasken-hpe: This code is exactly the same as lines 292-309, which suggests it could be refactored.

haasken-hpe (changes requested):
Needs to fix the mapping from `sat hwinv` section headers to the HSM types. Right now, it looks like the `--list-all` test cases don't really do anything because of this mismatch.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,416 @@
...
+            match = re.match(r"Listing of all (.+)", line.strip())
+            if match:
+                section_name = match.group(1)
+
+                # Check if the section corresponds to a component that exists in the system
+                component_type = None
+                for hsm_type, section in TYPE_TO_SECTION.items():
+                    if section == section_name:
+                        component_type = hsm_type
+                        break
```
haasken-hpe: This would actually be easier if the dictionary keys and values were swapped, so that it mapped from the `sat hwinv` section name to the HSM component type.

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
```diff
@@ -0,0 +1,416 @@
...
+                header_line_index = i + 3
+                if header_line_index >= len(stdout_lines):
+                    self.fail(f"Header row for section '{section_name}' not found in output.")
+                header_line = stdout_lines[header_line_index].strip()
+
+                # Parse the actual header
+                actual_header = [col.strip() for col in header_line.split("|") if col.strip()]
+
+                # Validate the actual header against the adjusted expected header
+                self.assertEqual( expected_headers, actual_header,
```
haasken-hpe:
```suggestion
                self.assertEqual(expected_headers, actual_header,
```

haasken-hpe (approved):

src/csm_testing/tests/sat_functional/hwinv/test_hwinv.py
haasken-hpe: This has been addressed with your latest changes.

haasken-hpe (approved):
Looks good!

PR #741, opened 2025-08-08
CRAYSAT-1968: create functional tests for sat status --format in csm-testing

ethanholen-hpe:
## Summary and Scope

Create functional tests for "sat status --format" in csm-testing

## Issues and Related PRs

* Resolves [CRAYSAT-1968](https://jira-pro.it.hpe.com:8443/browse/CRAYSAT-1968)

## Testing

### Tested on:

  * rocket
  * tyr

### Test description:

- Were the install/upgrade-based validation checks/tests run (goss tests/install-validation doc)? yes
- Were continuous integration tests run? If not, why? yes
- Was upgrade tested? If not, why? yes 
- Was downgrade tested? If not, why? yes 
- Were new tests (or test issues/Jiras) created for this change? no

## Risks and Mitigations

_Are there known issues with these changes? Any other special considerations?_ no

## Pull Request Checklist

- [x] Copyrights updated
- [x] License file intact
- [x] Target branch correct
- [x] CHANGELOG.md updated

### DIFF
--- a/src/csm_testing/tests/sat_functional/sat_testing_utils.py
+++ b/src/csm_testing/tests/sat_functional/sat_testing_utils.py
@@ -29,40 +29,36 @@ import sys
 import yaml
 
 
-class SatTestingUtils:
+def validate_json(json_string: str) -> json:
+    """
+    Validates a json string and returns the validated json
 
-    @staticmethod
-    def validate_json(json_string: str) -> bool:
-        """
-        Validates a json string
+    Args:
+        json_string (str): string containing json
 
-        Args:
-            json_string (str): string containing json
+    Returns:
+        json: a json object
+    """
+    try:
+        return json.loads(json_string)
+    except json.JSONDecodeError as err:
+        sys.stderr.write(f'Unable to decode json with error: {err}')
+        assert False, f"The provided string is not valid json: {json_string}"
 
-        Returns:
-            bool: True if json is valid false otherwise
-        """
-        try:
-            json.loads(json_string)
-            return True
-        except json.JSONDecodeError as err:
-            sys.stderr.write(f'Unable to load json with error: {err}')
-            return False
 
-    @staticmethod
-    def validate_yaml(yaml_string: str) -> bool:
-        """
-        Validates a yaml string
+def validate_yaml(yaml_string: str) -> yaml:
+    """
+    Validates a yaml string and returns the validated yaml
 
-        Args:
-            yaml_string (str): string containing yaml
+    Args:
+        yaml_string (str): string containing yaml.
 
-        Returns:
-            bool: True if yaml is valid false otherwise
-        """
-        try:
-            yaml.safe_load(yaml_string)
-            return True
-        except yaml.YAMLError as err:
-            sys.stderr.write(f'Unable to load yaml with error: {err}')
-            return False
+    Returns:
+        yaml: a yaml object.
+    """
+    try:
+        return yaml.safe_load(yaml_string)
+    except yaml.YAMLError as err:
+        sys.stderr.write(f'Unable to load yaml with error: {err}')
+        assert False, f"The provided string is not valid yaml: {yaml_string}"
+
--- a/src/csm_testing/tests/sat_functional/sat_testing_utils.py
+++ b/src/csm_testing/tests/sat_functional/sat_testing_utils.py
@@ -1,0 +1,68 @@
+#
+# MIT License
+#
+# (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and associated documentation files (the "Software"),
+# to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense,
+# and/or sell copies of the Software, and to permit persons to whom the
+# Software is furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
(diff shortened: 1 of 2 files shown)


ethanholen-hpe:
testing on rocket: https://gist.github.com/ethanholen-hpe/f9d2535c6bf0ba247702464ed591653a

src/csm_testing/tests/sat_functional/status/test_status.py
```diff
@@ -29,6 +29,7 @@
 import re
 from typing import List, Tuple
 import unittest
+from ..sat_testing_utils import  SatTestingUtils
```
haasken-hpe: I would suggest using an absolute import here. You should also group local app/library-specific imports into a separate group.

```suggestion

from csm_testing.tests.sat_functional.sat_testing_utils import SatTestingUtils
```

See: https://peps.python.org/pep-0008/#imports

src/csm_testing/tests/sat_functional/status/test_status.py
```diff
@@ -101,6 +111,17 @@ def test_status_command(self) -> None:
 
         self.assertEqual(expected_header, status_output_header)
 
+    def test_status_formatting_command(self) -> None:
```
haasken-hpe: I would suggest splitting into two separate tests. Otherwise, if the json test fails, the yaml test will not be finished.

src/csm_testing/tests/sat_functional/status/test_status.py
```diff
@@ -46,17 +47,26 @@
...
+    std_err = proc.stderr.decode().strip()
+
+    return std_output, std_err
+
+def get_header(command: str, num_lines_in_header: int) -> Tuple[str, str, str]:
+    """
+    Run the specified command.
+    Return the standard output, standard error, and the header (meaning the first
+    num_lines_in_header lines in the output)
+    """
```
haasken-hpe: I would suggest updating both `get_header` and `get_command_output` to use google-style Python docstrings for consistency.

src/csm_testing/tests/sat_functional/sat_testing_utils.py
```diff
@@ -0,0 +1,68 @@
...
+#
+"""
+Defines utility functions used across the sat_functional tests
+"""
+import json
+import sys
+import yaml
+
+
+class SatTestingUtils:
```
haasken-hpe: I don't see a lot of benefit other than just namespacing to putting these functions as static methods in a class. Did you have some reason in mind? Otherwise they could just be functions in this module.

One way you could get some benefit is if you made this a mixin class that could use `self.assert` methods to assert valid JSON and fail with a message about bad JSON/YAML if it fails to load as JSON/YAML. If you did that, you could also return the loaded object for further validation by the caller if they wanted to do that.

src/csm_testing/tests/sat_functional/sat_testing_utils.py
```diff
@@ -0,0 +1,68 @@
...
+            json_string (str): string containing json
+
+        Returns:
+            bool: True if json is valid false otherwise
+        """
+        try:
+            json.loads(json_string)
+            return True
+        except json.JSONDecodeError as err:
+            sys.stderr.write(f'Unable to load json with error: {err}')
```
haasken-hpe: We are using the `logging` module in `test_bootprep`, so you could probably do that here, but it wouldn't be necessary if you made this just fail with an assertion (either from `unittest.TestCase` assertion methods or just a plain `assert`).

src/csm_testing/tests/sat_functional/status/test_status.py
```diff
@@ -101,6 +111,17 @@ def test_status_command(self) -> None:
...
 
+    def test_status_formatting_command(self) -> None:
+        """Test that `sat status --format (json | yaml)` returns the properly formatted response"""
+        json_command = 'sat status --format json'
+        yaml_command = 'sat status --format yaml'
+
+        json_std_out, json_std_err = get_command_output(json_command)
+        yaml_std_out, yaml_std_err = get_command_output(yaml_command)
+
+        self.assertTrue(SatTestingUtils.validate_json(json_std_out))
```
haasken-hpe: It wouldn't hurt to do some validation of the data beyond just that it is valid JSON/YAML. You could check just the fields as the other tests here do. In order to do so, I think you'd have to check the keys of the first element of the array in the loaded JSON since there is no explicit header in JSON/YAML output. You just get an array of components with the keys being the headings.

haasken-hpe (commented):
Some minor suggestions to consider.

ethanholen-hpe:
Testing on Tyr: https://gist.github.com/ethanholen-hpe/5d5667cba5671bab7fd3d16aac7dd406

src/csm_testing/tests/sat_functional/status/test_status.py
```diff
@@ -30,6 +30,8 @@
 from typing import List, Tuple
 import unittest
 
+from csm_testing.tests.sat_functional.sat_testing_utils import *
```
haasken-hpe: It's a bad practice to do `from ... import *` because then it can be hard to trace where symbols are coming from. Just update to import the two functions you need.

src/csm_testing/tests/sat_functional/status/test_status.py
```diff
@@ -101,6 +121,42 @@ def test_status_command(self) -> None:
...
+        json_response = validate_json(json_std_out)
+
+        self.assertTrue(json_response)
+
+        json_header_keys = list(json_response[0].keys())
+
+        status_command = 'sat status'
+        status_output, status_err, status_output_header_str = get_header(status_command, 3)
+        status_all_output = status_output + '\n' + status_err
+        expected_header = adjust_expected_header(SAT_STATUS_HEADER, status_all_output)
```
haasken-hpe: I'm not sure I understand why you have to call `sat status` again to get the expected header lines. Can you explain that? Is it possible to achieve this with just the `json_std_err`? Or do we not omit missing/empty fields when the user specifies JSON or YAML format?

haasken-hpe (approved):
Looks good, but I would change the import and consider if you can do without the second call to `sat status` in each test to construct the expected header.

src/csm_testing/tests/sat_functional/status/test_status.py
haasken-hpe:
```suggestion
        expected_header = adjust_expected_header(SAT_STATUS_HEADER, json_std_err)
```

ethanholen-hpe:
Final testing on Vinland: https://gist.github.com/ethanholen-hpe/5d8d96724ff835d14157af4fdeacb987

